PR #3, opened 2016-06-03
archweb netboot and sponsorship updates

brain0:


templates/public/download.html
```diff
@@ -68,8 +68,8 @@
 
     <p>If you have a wired connection, you can boot the latest release directly over the network.</p>
     <ul>
-        <li><a href="{{ releng_pxeboot_url }}"
-            title="Arch Linux Netboot Live System">Arch Linux Netboot</a></li>
+        <li><a href="/releng/netboot/"
+            title="Arch Linux Netboot">Arch Linux Netboot</a></li>
```
angvp: Please use url tag and the name that you gave to that url

https://docs.djangoproject.com/es/1.9/ref/templates/builtins/#url

releng/urls.py
```diff
@@ -23,7 +23,8 @@
 )
 
 netboot_patterns = patterns('releng.views',
-    (r'^archlinux\.ipxe$', 'netboot_config', {}, 'releng-netboot-config')
+    (r'^archlinux\.ipxe$', 'netboot_config', {}, 'releng-netboot-config'),
+    (r'^$',                'netboot_info',   {}, 'releng-netboot-info')
```
angvp: ^ this is anti pep8 please attach to the pep8 guideline as much as possible

PR #8, opened 2017-03-20
Define BOOTIF in extrabootoptions

thurstylark:
Since this script sets 'ip=dhcp' by default, archiso_pxe_common will fail if more than one network interface exists, and at least one interface cannot be configured by dhcp. Setting BOOTIF will define which interface that archiso_pxe_common should configure, and will cause it to ignore the rest.

Fixes https://bugs.archlinux.org/task/50448

### DIFF
--- a/templates/releng/archlinux.ipxe
+++ b/templates/releng/archlinux.ipxe
@@ -9,7 +9,7 @@ imgtrust
 
 # initial options
 set release {{ releases.0 }}
-set extrabootoptions ip=dhcp net.ifnames=0
+set extrabootoptions ip=dhcp net.ifnames=0 BOOTIF=01-${netX/mac}
 set countrycode
 
 :main


angvp:
@brain0 would you mind to help me reviewing this?

thurstylark:
I'm willing to answer any questions if necessary. I would like to have this ready for next month's ISO build if possible.

brain0:
The netboot options are independent of the ISO builds, they are enabled as soon as the website is updated.

As for the change itself, I had no chance to test it yet and don't know wether BOOTIF has the desired effect. Other than that, the change looks fine. If this works, this is definitely something we want. Since the argument is added to extrabootoptions, the user can still change this option before booting anyway.

thurstylark:
Ahh, right. Forgot this script doesn't get built in.

BOOTIF is documented along with the other options for the archiso_pxe_common hook[1]. BOOTIF is usually set by pxelinux for the same purpose.

[1]: https://git.archlinux.org/archiso.git/tree/docs/README.bootparams#n72

PR #9, opened 2017-03-22
Replace SubfieldBase as it is deprecated

MarcoCLA:
SubfieldBase is deprecated in upcoming versions of Django


grazzolini:
This work was already done by Jelle:
https://github.com/jelly/archweb

I did preliminary tests on his branch it is working. You could help improving from there, to avoid duplicate efforts.

PR #10, opened 2017-04-02
Use x86_64 instead of dual postfix for ISO image names

pierres:
Fixes FS#53524 and FS#53520


angvp:
Thanks for the PR @pierres I'm merging this today with the other opened prs

PR #11, opened 2017-04-17
Add travisci file

jelly:


angvp:
Hey @jelly instead of doing the cp local_settings.example.py should be better to refactor settings.py with this:

    try:
        import local_settings
    except ImportError:
        pass

So it won't break and no need to do that cp :)

jelly:
With the try / except ImportError collecstatic does not work.

```
(env) [jelle@lenovomate][~/projects/archweb]%python2 manage.py collectstatic --noinput
Unknown command: 'collectstatic'
```

And secret key is not imported:
```
    raise ImproperlyConfigured("The SECRET_KEY setting must not be empty.")
django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.
```

angvp:
so we should set a bogus empty key then, and local_settings will override it.

PR #18, opened 2017-05-12
mirrors: add simple test for mirror stats

jelly:
Add a simple /mirrors/status test.

\o/ more tests 👍

### DIFF
--- a/mirrors/tests.py
+++ b/mirrors/tests.py
@@ -1,0 +1,62 @@
+import json
+
+from django.test import TestCase
+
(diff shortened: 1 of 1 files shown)


angvp:
@jelly tests are not passing, :P can you correct this and I will merge right away :) thanks!

jelly:
@angvp I'll look at it. It's strange that it fails.

jelly:
So it fails since we have a decorator "cache_function" which caches the first instance of the data for the "view". So that's kinda annoying.

PR #22, opened 2017-05-22
releng: Fix django warnings

jelly:
Fixes warnings for releng.Test.modules and releng.Test.rollback_modules
(fields.W340) null has no effect on ManyToManyField.


jelly:
@angvp am I forgetting to commit a mirgration here btw?

angvp:
@jelly not sure if you miss a migration if the command django makemigrations didn't make it one then it's ok

jelly:
@angvp Yup I forgot to run makemigrations, I've squashed the migrations into the commit.

PR #33, opened 2017-06-11
Tests add 12% more coverage

jelly:
Create a coveragerc file, fix a faulty test, completely test visualize/*. Add missing feed and url tests.


jelly:
Seems I lied, but 5.2% it is :)

jelly:
And by excluding travisci, it's now 52% the same as locally :)

PR #36, opened 2017-06-13
Filter the users that appear on the devel clocks page

grazzolini:
Since we have testers and other kind of users, filter the users that
appear on the devel clocks page for membership on Developers, Trusted Users
or Support Staff only.

### DIFF
--- a/devel/views.py
+++ b/devel/views.py
@@ -7,10 +7,10 @@ from django.contrib.admin.models import ADDITION, LogEntry
 from django.contrib.auth.decorators import (login_required,
                                             permission_required,
                                             user_passes_test)
-from django.contrib.auth.models import User
+from django.contrib.auth.models import Group, User
 from django.contrib.contenttypes.models import ContentType
 from django.db import transaction
-from django.db.models import Count, Max
+from django.db.models import Count, Max, Q
 from django.http import Http404, HttpResponseRedirect
 from django.shortcuts import get_object_or_404, render
 from django.utils.encoding import force_unicode
@@ -107,8 +107,9 @@ def stats(request):
 
 @login_required
 def clock(request):
-    devs = User.objects.filter(is_active=True).order_by(
-        'first_name', 'last_name').select_related('userprofile')
+    devs = User.objects.filter(is_active=True).filter(groups__in=Group.objects.filter(
+        Q(name='Developers') | Q(name='Trusted Users') | Q(name='Support Staff'))).order_by(
+        'first_name', 'last_name').select_related('userprofile').distinct()
 
     latest_news = dict(News.objects.filter(author__is_active=True).values_list(
         'author').order_by().annotate(last_post=Max('postdate')))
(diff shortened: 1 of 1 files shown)


angvp:
Have you tested the performance of that query? please grab the production database and test it on  your environment, I'm afraid that query can overload the server and we should write it different

grazzolini:
I've just ran it directly on apollo (using manage.py) and it took less than 0.001 second. I know the query is not optimal, but I think we would need to have far more users to make it an issue. Just as a comparison, I had a DB with more than 100k users and ran much more complex queries on them using Q with no (noticeable) performance issues on inferior hardware than apollo's.

The other option here, is to use a raw query. I can write one, if you want.

devel/views.py
```diff
@@ -107,8 +107,9 @@ def stats(request):
 
 @login_required
 def clock(request):
-    devs = User.objects.filter(is_active=True).order_by(
-        'first_name', 'last_name').select_related('userprofile')
+    devs = User.objects.filter(is_active=True).filter(groups__in=Group.objects.filter(
+        Q(name='Developers') | Q(name='Trusted Users') | Q(name='Support Staff'))).order_by(
+        'first_name', 'last_name').select_related('userprofile').distinct()
```
angvp: Instead of using 3 times Q I rather I will filter them with a python list comprehension

```
SELECTED_GROUPS = ['Developers', 'Trusted Users', 'Support Staff']


    def clock(request):
        groups = Group.objects.all().only('id', 'name')
        groups_ids = [x.id for x in groups if x.name in SELECTED_GROUPS]
        users = User.objects.filter(is_active=True).filter(
            groups__in=groups_ids).select_related('userprofile').distinct()
```

I'd go with doing an implementation like this ^ for two reasons, despite if it's optimized or not (I don't think either this version is actually a super optimized version):

1.- Legibility (Reading between all Q queries mixed with the rest of the queries might lead to confusion), and a raw query will be even worst for that.
2.- Maintainability (just adding/removing another value to the list SELECTED_GROUP is easier than adding a Q condition with the high chances of screwing it everything)

grazzolini:
My first approach was using a list and comprehension ;-). But I did not want to affect our coverage and also keep the code mostly the same. I think you don't need that .all() query first, you can do it using a filter and _in. Let me see if I can simplify it further. We don't have that many groups, but an all() seems unnecessary to me.

angvp:
@grazzolini I agree both approaches work well, but still, let me try to explain the difference between both approaches

approach a): 1 query of the type:
```
 SELECT id, name FROM foo (in a table of 3 results is just one small network operation)
```
And then filter the resultset with python, and send just the indexes for the second query

approach b) 1 query of the type
```
SELECT * FROM foo WHERE bar IN ('x', 'y', 'z')

And then send the whole resultset to the Django ORM to fire the second query
```
Note to mention that this query fires a lookup on that table when running the WHERE clause on the field bar (and any value inside the IN) because IIRC that field doesn't have an index, so, full table scan?

I think approach a) is quite better, even though we cant see any performance difference because is too minimal or it might be just nanoseconds  (because this table just have 3 records).

I'm ok to merge this because as I said both approaches work and there is no a performance difference, but wanted to explain my point. :-)

grazzolini:
I agree with you. I've always found Groups on django being one of it's poorest implemented things. You need to filter on permissions or using group names. Having said that, we can create a migration that makes the group name, to be a db index. If you want I can rework the query further.

PR #38, opened 2017-06-22
Change navbar css so that links wrap in header on narrow windows

tyzoid:
Previously, all of these links would drop beneath the header,
making the navigation difficult to use, as it would become white
links on an off-white background.


angvp:
@jelly wanna review this? I don't see any difference?

PR #42, opened 2017-06-28
Add setup.cfg with pytest configuration

blueyed:
This allows for using pytest easily, after installing it (and
pytest-django).

TODO:
- [ ] update after/for https://github.com/archlinux/archweb/pull/48#issuecomment-312925869

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,6 @@
+[tool:pytest]
+DJANGO_SETTINGS_MODULE = settings
+
+# Cope with tests.py files in separate subdirs.
+python_files = tests.py
+testpaths = devel main mirrors news packages public releng retro visualize


jelly:
Any reason to prefer pytest above nose?

blueyed:
Currently it is just a wrapper, using the Django-compatible mode, but I would really like it to become the default test runner, so that its features (https://docs.pytest.org/en/latest/#id1) could be used.
I do not want to push for that, but this config allows for using it with the existing unittest/Django based tests already for now.

angvp:
As long as I'd like to use pytest eventually I don't see yet this PR as it will be merged now, thanks anyway @blueyed I will let it open for later when refactoring some of the test architecture.

jelly:
I've chosen to prefer a pytest.ini in https://github.com/jelly/archweb/commit/bda9a1d197ba70fa9d638a3ca195ece41dfe1516

PR #48, opened 2017-07-01
devel: Add reporead test

jelly:
Add a test for reporead import database functionality. Move tests for
devel to a separate directory and depend on mock for patching the
logging functonality. (Included by default in Python 3)


devel/tests/test_user.py
```diff
@@ -1,8 +1,8 @@
 from django.contrib.auth.models import User
 from django.test import TestCase
 
-from .utils import UserFinder
-from .models import UserProfile
+from devel.utils import UserFinder
+from devel.models import UserProfile
```
blueyed: I appreciate the renaming here, but why?
Needs to be covered in https://github.com/archlinux/archweb/pull/42.

blueyed:
btw: https://github.com/pytest-dev/pytest-mock/

devel/tests/test_user.py
jelly: Without it, the PR will obviously not work and we want the tests moved in subdiir/tests/test_component.py

jelly:
@blueyed that's an external dep, Python 3 has a builtin unittest.mock which is way more appealing https://docs.python.org/3.5/library/unittest.mock.html

jelly:
@angvp Resolved the merge commits.

PR #49, opened 2017-07-09
Upgrade js

jelly:
Note this is potentially breaking, but our currenty JQuery version is 5 years old. I've hacked gotten typeahead to work and the table sorter. I feel it needs some more testing with more appropriate testing data.

I have a side branch to upgrade d3.js but it actually broke (a newer version).


jelly:
And now upgraded to 3.2.1, seems to also work. I might want to squash that commit into Update JQuery.

angvp:
@jelly this is the next big refactor I'm about to merge soon (as it might break stuff and we need to test it carefully), would you like to squash it and we merge/deploy this after we tested properly? :)

PR #51, opened 2017-07-15
Fix the HTTP error 500 when trying to access the admin_log page.

grazzolini:
When trying to access the admin_log page from Django admin, there
was missing the request from the each_context call.


grazzolini:
We should use #52 in favor of this one. Jelle already made a test for it

PR #55, opened 2017-08-14
Donate

grazzolini:
Sponsor cleanup

### DIFF
--- a/templates/public/donate.html
+++ b/templates/public/donate.html
@@ -36,15 +36,7 @@
 
     <h3>Commercial sponsors and contributions</h3>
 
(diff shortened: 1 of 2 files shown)


jelly:
Test fails since the logo is still used in the "retro" pages. So either remove it there or re-add it.

grazzolini:
Not just that, the logo is on the index page as well. I forgot to search for it. I need to remind myself to always run the tests before pushing.

jelly:
You might wanna squash the commits into one commit 👍

grazzolini:
Instead of squash, I marked both as fixup, so they don't appear on the log. It is as if they did not happen.

PR #57, opened 2017-08-28
Make pylint scan all module directory's

jelly:
Instead of only scanning the base *.py files, scan all module
directories.


jelly:
This Pr enables some more pylint checks, for example the deprecated module one is useful for removing old code. Also speed up the travisci builds a bit by caching PIP packages.

PR #58, opened 2017-08-28
Added removed rss

jelly:
This PR adds two feeds, one for newly added packages (which did not exists in the repo before) and removed packages. This requires triggers to be configured in your database for updates. For SQLite they are not setup for some reason by default? Is it possible we can make this happen? @angvp @grazzolini? Then we can at least right some tests.

The triggers can be found in packages/sql. packages/sql/update.sqlite3.sql for sqlite and packages/sql/update.postgresql_psycopg2.sql for PostgreSQL which seems to be enabled and working in production! 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,13 @@ packages, you will probably want the following:
 Alter architecture and repo to get x86\_64 and packages from other repos if
 needed.
 
+8. Database Updates for Added/Removed packages
+
+        sqlite3 archweb.db < packages/sql/update.sqlite3.sql
+
+For PostgreSQL use packages/sql/update.postgresql_psycopg2.sql
+
+
 # Testing SMTP server
 
 To be able to create an account on your test environment an SMTP server is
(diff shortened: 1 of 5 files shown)


angvp:
@jelly I was reviewing this again, we briefly talked about this PR before, I think we might need to update this to not reflect unsupported architectures, also the triggers for the sql are not on this pr and that was one of the main stoppers iirc?

jelly:
@angvp The triggers are actually setup on our live env, but not enabled in the sqlite env. The triggers are located in packages/sql/ but if I recall correctly @grazzolini wanted to change this to a Django specific "trigger" of which I forgot the name.

grazzolini:
@jelly 
I wanted us to use data migrations for this, because they are DB agnostic and you can even write functions to undo the changes, if wanted:

https://docs.djangoproject.com/en/2.0/topics/migrations/#data-migrations

That's for django 2.0.

PR #62, opened 2017-10-11
donate: add Icons8

Barthalion:


templates/public/donate.html
```diff
@@ -53,6 +53,15 @@
     <img width="496" height="76" src="{% static "pia_logo.png" %}"
     class="sponsor-btn-pia" title="" alt="Private Internet Access logo"/></a>
 
+    <p>We would also like to thank <a href="https://icons8.com/" title="Icons8">
+    Icons8</a> for their yearly monetary donation. Icons8 is a team of talented
+    designers and engineers working on fullfilling all your icons needs.</p>
```
polyzen: fulfilling*

Barthalion: There is no such thing as too many l's!

Thanks, will fix it.

jelly:
Looks good to me 👍

templates/public/donate.html
Barthalion: Done.

PR #63, opened 2017-10-24
package details: remove category from bugtracker links

eli-schwartz:
This completely screws up searching for bugs, as you can easily miss any
bugs filed to the testing package, and will *always* miss the ones that
have been moved to "Upstream Bugs".

There is no purpose in restricting the view of bugs filed against a
specific package, to some arbitrary subcategory of "why" the bug was
filed; all it accomplishes is causing unknowing users to file duplicate
bug reports.

### DIFF
--- a/main/templatetags/details_link.py
+++ b/main/templatetags/details_link.py
@@ -32,7 +32,6 @@ def bugs_list(package):
     url = "https://bugs.archlinux.org/"
     data = {
         'project': package.repo.bugs_project,
-        'cat[]': package.repo.bugs_category,
         'string': package.pkgname,
     }
     return link_encode(url, data)


Scimmia22:
For the record, this is a change I've wanted since I started as a bug wrangler. toofishes was maintaining ArchWeb at the time and didn't want to change it as he thought the link should show tickets that needed to be dealt with, not ones that have already been "wrangled". Even as-is, it didn't accomplish that goal, so changing this has no drawbacks and many real benefits.

eli-schwartz:
> thought the link should show tickets that needed to be dealt with

Huh, I'd think showing tickets in any category, as long as their status was Unconfirmed, would do a better job there. :grin:

Anyway, the purpose of the link is to help users discover whether there are currently-open bugs with the package (which might impact them), the only ones who care about "unwrangled" bugs are me and @Scimmia22 and we don't generally work our way from the package details to the bugtracker...

PR #64, opened 2017-11-07
Drop i686

jelly:
Remove mention of i686 on the public parts of the website.

### DIFF
--- a/templates/public/feeds.html
+++ b/templates/public/feeds.html
@@ -24,13 +24,7 @@
 
     <p>If you are interested in <a href="/feeds/packages/" class="rss"
         title="Arch Linux package updates feed">all package updates</a>,
-    then grab this feed. Note that when a package is updated for multiple
-    architectures, you will see each individual update show up here.
-    Alternatively, you can select a packages feed from the below table that is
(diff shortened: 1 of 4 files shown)


templates/public/index.html
```diff
@@ -16,8 +16,8 @@
     lightweight and flexible Linux® distribution that tries to Keep It
     Simple.</p>
 
-    <p>Currently we have official packages optimized for the i686 and
-    x86-64 architectures. We complement our official package sets with a
+    <p>Currently we have official packages optimized for the x86-64
+    architectures. We complement our official package sets with a
```
Barthalion: architecture

templates/public/feeds.html
```diff
@@ -29,7 +29,7 @@
     Alternatively, you can select a packages feed from the below table that is
```
Barthalion: I'd rather just remove entire block about architecture specific feeds given we will have only one for now.

angvp: Agreed

jelly: Hmm well we will still show "any" / "x86_64" and we will still show "All arches". I believe we should keep this since in the future we might want to support a new Arch.

PR #67, opened 2017-11-26
Return different dep types under different keys in JSON results

kyrias:
Currently all dependencies are under the same `depends` key in the JSON results.  With this commit they're instead sorted under keys for each dependency type instead.

### DIFF
--- a/packages/utils.py
+++ b/packages/utils.py
@@ -472,6 +472,9 @@ def get_signoff_groups(repos=None, user=None):
     return signoff_groups
 
 
+DEPENDENCY_TYPES =  [('D', 'depends'), ('O', 'optdepends'),
+                     ('M', 'makedepends'), ('C', 'checkdepends')]
+
 class PackageJSONEncoder(DjangoJSONEncoder):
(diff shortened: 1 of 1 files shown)


Barthalion:
Actually, it would make sense to put that also for packages json view.

kyrias:
Both use the PackageJSONEncoder.

Barthalion:
:+1:

packages/utils.py
```diff
@@ -488,6 +488,10 @@ def default(self, obj):
             data = {attr: getattr(obj, attr) for attr in self.pkg_attributes}
             for attr in self.pkg_list_attributes:
                 data[attr] = getattr(obj, attr).all()
+            all_deps = obj.depends.all()
+            for (deptype, name) in [('D', 'depends'), ('O', 'optdepends'),
+                                    ('M', 'makedepends'), ('C', 'checkdepends')]:
```
angvp: Just a coding style thing, I'd use a constant called DEPENDENCIES_TYPE with that list, and iterate that instead

PR #69, opened 2017-12-03
Bump results limit of search json view to 1000

Barthalion:
This is for my completely selfish reasons, but it makes no sense to limit the results while there's no pagination for it.

### DIFF
--- a/packages/tests.py
+++ b/packages/tests.py
@@ -67,7 +67,6 @@ class PackageSearchJson(TestCase):
         response = self.client.get('/packages/search/json/')
         self.assertEqual(response.status_code, 200)
         data = json.loads(response.content)
-        self.assertEqual(data['limit'], 250)
         self.assertEqual(data['results'], [])
         self.assertEqual(data['valid'], False)
 
--- a/packages/views/search.py
+++ b/packages/views/search.py
@@ -1,6 +1,7 @@
 import json
 
 from django import forms
+from django.conf import settings
 from django.contrib.auth.models import User
 from django.db.models import Q
 from django.http import HttpResponse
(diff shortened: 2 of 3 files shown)


jelly:
I'm not sure how upping the limit affect archweb's performance and load on the postgres db.

Barthalion:
I applied the change on production, and so far, I see no significant changes in Grafana.

angvp:
@Barthalion Do you mind if, we move this value to a setting and import it from there, so in the future change this value is something that can be done w/o touching code, but just a setting on re run ansible?

Barthalion:
Done.

angvp:
@Barthalion tests are failing now, can you check this?

And please when importing vars from settings do it this way:
 
    from django.conf import settings

    JSON_RESULTS_LIMIT = getattr(settings, 'JSON_RESULTS_LIMIT', 1000)

And do not use settings.JSON_RESULTS_LIMIT directly this might lead to future breakages when handling different settings file (local_settings etc)

kyrias:
So to solve this the more "proper" way I threw together a quick pagination thing (https://github.com/kyrias/archweb/commit/cecfba83ae0ef760396a471deab61a6d318c55df) using `django.core.paginator`, and it seems to be working properly from some quick tests.  Any comments/thoughts about it?

I haven't looked at any tests and such for it, but they should really just keep working as-is.

jelly:
That commit looks neat, but shouldn't we define a global limit from where you need to fetch more pages?

angvp:
I like more the initiative of @kyrias when we merge #73 then I'll close this PR, thanks for the effort tho @Barthalion :)

Barthalion:
Either way works, I really just needed this for my orphan status script and I complained about lack of pagination on IRC when I bumped the limit.

PR #70, opened 2017-12-28
Add admin emails to the mirror status JSON output when logged in

kyrias:


mirrors/views/api.py
```diff
@@ -92,6 +92,9 @@ def mirror_details_json(request, name):
             show_all=authorized)
     data = status_info.copy()
     data['version'] = 3
+    if authorized:
+        data['admin_email'] = mirror.admin_email
+        data['alternate_email'] = mirror.alternate_email
```
jelly: Do we want to restrict this to only specific roles? @Bluewind

Bluewind: That would be better, yes. Can we restrict it the same way the email is restricted on the html version?

jelly: We can probably restrict it on either staff groups or user permissions

Bluewind: Looks like we currently make this public to all logged in users[1]. I'd prefer having it check for the mirrors.change_mirror permission, but it's up to you. 

[1] https://github.com/archlinux/archweb/blob/master/templates/mirrors/mirror_details.html#L35

kyrias: Done

Bluewind:
Good from my side

jelly:
Thanks for the PR, merged :)

PR #71, opened 2017-12-28
Add mirror detail URLs to JSON output

kyrias:


mirrors/views/api.py
```diff
@@ -92,6 +93,7 @@ def mirror_details_json(request, name):
             show_all=authorized)
     data = status_info.copy()
     data['version'] = 3
```
jelly: Hm stupid question, is this the data version and should be bump it :)

kyrias: Hm, no clue actually.  I guess it would make sense to bump it, though on the other hand very few people would see any difference.  Up to you!

PR #73, opened 2017-12-30
Add pagination to JSON package search

kyrias:
https://github.com/archlinux/archweb/pull/69


main/fixtures/package.json
```diff
@@ -1,25 +1,132 @@
...
+      "installed_size": 18100224,
+      "packager_str": "Dave Reisner <dreisner@archlinux.org>",
+      "last_update": "2017-12-30T02:21:43.865Z",
+      "arch": 3,
+      "packager": null,
+      "pkgver": "232"
+    },
+    "model": "main.package",
+    "pk": 5
+  }
```
kyrias: I wrote a small script to `reporead` the `devel/fixtures/core.db.tar.gz` database and then use Django to dump the package fixture again, running it through `jq` for formatting.  Might be useful to add this to the repo as well?

kyrias:
Anyway, any comments/critiques welcome.

We could also potentially have 250 or whatever as the default limit but also allow a higher one, but not sure whether that'd be worth it.

packages/views/search.py
```diff
@@ -14,6 +15,8 @@
 
 
 class PackageSearchForm(forms.Form):
+    limit = forms.CharField(required=False)
+    page = forms.CharField(required=False)
```
kyrias: Would be nice for these to be proper `IntegerField`s, but they cannot have a default value, so even if you mark them as not required we'll get an exception raised when we try to use the value.

Additionally since we're on an old version we can't specify a default value for `CharField`, but once archweb is finally upgraded to newest Django we could change that to have the defaults specified here.

main/fixtures/package.json
jelly: Hmm yeah, but I'm thinking that we can better get rid of the core.db.tar.gz fixture and just generate the test data.

packages/views/search.py
jelly: Too bad, but hopefully we will update to a more modern django this year.

main/fixtures/package.json
kyrias: I think it would likely make sense to have a JSON or whatever format for specifying the same data as in the DB file, but contained in a single file that can be easily diffed and such, and then have a function à la `reporead` that would insert the data into the archweb DB.

PR #75, opened 2018-01-02
Add opensearch tests

jelly:
Add additional tests for our opensearch endpoints

### DIFF
--- a/packages/tests.py
+++ b/packages/tests.py
@@ -163,4 +163,20 @@ class PackageSearch(TestCase):
         self.assertEqual(response.status_code, 200)
         self.assertIn('5 matching packages found', response.content)
 
+class OpenSearch(TestCase):
+    fixtures = ['main/fixtures/arches.json', 'main/fixtures/repos.json',
+                'main/fixtures/package.json']
+
+    def test_packages(self):
+        response = self.client.get('/opensearch/packages/')
+        self.assertEqual(response.status_code, 200)
+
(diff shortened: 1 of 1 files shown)


angvp:
@jelly I'd keep the naming scheme pythonic test_packages and test_packages_suggest for the functions instead camel case names, also, not sure if the test runner takes these?

packages/tests.py
```diff
@@ -163,4 +163,21 @@ def test_filter_packager_unknown(self):
         self.assertEqual(response.status_code, 200)
         self.assertIn('5 matching packages found', response.content)
 
+class OpenSearch(TestCase):
+    fixtures = ['main/fixtures/arches.json', 'main/fixtures/repos.json',
+                'main/fixtures/package.json']
+
+
+    def testPackages(self):
```
angvp: test_packages instead

packages/tests.py
```diff
@@ -163,4 +163,21 @@ def test_filter_packager_unknown(self):
...
+        response = self.client.get('/opensearch/packages/')
+        self.assertEqual(response.status_code, 200)
+
+    def testPackagesSuggest(self):
```
angvp: test_packages_suggest

packages/tests.py
jelly: Fixed

packages/tests.py
jelly: fixed

PR #81, opened 2018-02-14
Add donation import management script

jelly:
This script imports messages from a Maildir folder and creates new
unique Donors based on the name in the subject. The subject also
contains the amount and the email address, which is not stored by
Archweb.

The input has to stripped quiet a bit, some entries are fully capitalised for example:

Receipt [$25.00] By: JOHN DOE [john.doe@archlinux.org]

Or contain extra numbers:

Receipt [$25.00] By: JOHN DOE 234234324 [john.doe@archlinux.org]

Or contain no name at all:

Receipt [$25.00] By: - - [john.doe@archlinux.org]

I've skipped / edited these entries

### DIFF
--- a/main/management/commands/donor_import.py
+++ b/main/management/commands/donor_import.py
@@ -1,0 +1,121 @@
+# -*- coding: utf-8 -*-
+"""
+donor_import
+
+Imports donators from the emails which are send to donate@archlinux.org,
+the subject of them email contains the name of the donator, the amount and
+the email address. Archweb's Donate model only contains the name, which is
+unique.
+
+An example subject:
+
+Subject: Receipt [$25.00] By: John Doe [john.doe@archlinux.org]
+
+Usage: ./manage.py donor_import path/to/maildir/
+"""
+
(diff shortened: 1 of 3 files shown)


Bluewind:
I'm not really into python so I can't spot subtle bugs, but the general idea/code looks good.

One issue I'm interested in is how this will react to emails that do not follow the expected format. Ones that do not contain "By:" or "[". Does it skip them or might it add things we don't actually want, like for example the whole subject?

jelly:
@Bluewind hmmm, I was planning on writing some tests, this might indeed fall through.

Might be easier to depend on the "parse" module and use that for parsing:

```
>>> parse("Receipt [{amount}] By: {item} [{email}]", "Receipt [$25.00] By: woop [john.doe@archlinux.org]")
<Result () {'amount': '$25.00', 'item': 'woop', 'email': 'john.doe@archlinux.org'}>
```

jelly:
The latest commit, now also checks the structure.

PR #85, opened 2018-02-18
mirrors/views/api: Add URL active status

kyrias:


Bluewind:
I'll assume this works fine for logged in users, but what do normal people see? Do they just get the same list as before only with the active field always being true?

kyrias:
Yeah.

PR #86, opened 2018-03-13
Fix release details list

jelly:


jelly:
@angvp this restores the accidently removed release details list and I've restored the link to it.

PR #88, opened 2018-03-18
Upgrade django

jelly:
Here we go, finally the update to the latest Django, this needs might still contains upgrade issues so testing is appreciated!


angvp:
great job @jelly !!! the coverage decreased :( we should try to add some more tests later (is a huge decrease) from 40.5 to 24.5 .. maybe in the port to python3 and django2 we should revisit some of these tests..

That said, I will try the website this week, if the manual tests passed I'd say to do the migration on next weekend

jelly:
@angvp That's strange that it has decreased so much, I'll look into it, since I always want to redo the commits a bit, since I have several commits bumping the django version :)

We have a staging env on https://archweb-dev.archlinux.org/ but it needs to be updated as well.

angvp:
@jelly right, I think they don't use the ```tests.py``` file anymore but rather something like ```tests/test_file.py```

PR #92, opened 2018-03-25
Extended tests

jelly:
Add / extend some new tests to further improve coverage and maybe even uncover bugs in the django upgrade branch.


angvp:
Check travis output @jelly

jelly:
@angvp fixed, it was that damn django caching :p

PR #101, opened 2018-04-24
Python 3 migration fixes and removing false positives from tests

twelveeighty:
Updated the README to be in line with Python 3 usage
Fixed a number of Python 3 issues due to encoding
Fixed a chunk of false positives generated during the automated tests due to either Django bug (https://code.djangoproject.com/ticket/10827) or changed 'expected' Python 3 error message.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ env/
 
 # rope
 .ropeproject/
+
+# coverage
+.coverage
--- a/README.md
+++ b/README.md
@@ -17,8 +17,8 @@ See AUTHORS file.
 
 # Dependencies
 
-- python2
-- python2-virtualenv
+- python
+- python-virtualenv
 
 # Python dependencies
 
@@ -27,15 +27,15 @@ use virtualenv and pip to handle these. But if you insist on (Arch Linux)
 packages, you will probably want the following:
 
 - django
-- python2-psycopg2
-- python2-markdown
-- python2-memcached
+- python-psycopg2
+- python-markdown
+- python-memcached
 
 # Testing Installation
 
-1. Run `virtualenv2`.
+1. Run `virtualenv3`.
 
-        cd /path/to/archweb && virtualenv2 ./env/
+        cd /path/to/archweb && virtualenv3 ./env/
 
 2. Activate the virtualenv.
 
@@ -81,9 +81,9 @@ needed.
 To be able to create an account on your test environment an SMTP server is
 required. A simple debugging SMTP server can be setup using Python.
 
-        python2 -m smtpd -n -c DebuggingServer localhost:1025
+        python -m smtpd -n -c DebuggingServer localhost:1025
 
-In local_settings.py change the EMAIL_HOST to 'localhost' and the EMAIL_PORT to
+In local\_settings.py add entries to set EMAIL\_HOST to 'localhost' and EMAIL\_PORT to
 1025.
 
 # Running tests and coverage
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -13,3 +13,6 @@ env/
 
 # rope
 .ropeproject/
+
+# coverage
+.coverage
```
jelly: This is an unrelated change, but sure I'll merge it..

README.md
```diff
@@ -81,9 +81,9 @@ needed.
 To be able to create an account on your test environment an SMTP server is
 required. A simple debugging SMTP server can be setup using Python.
 
-        python2 -m smtpd -n -c DebuggingServer localhost:1025
+        python -m smtpd -n -c DebuggingServer localhost:1025
 
-In local_settings.py change the EMAIL_HOST to 'localhost' and the EMAIL_PORT to
+In local\_settings.py add entries to set EMAIL\_HOST to 'localhost' and EMAIL\_PORT to
```
jelly: I don't see why this is required? It renders fine on Github.

main/management/commands/donor_import.py
```diff
@@ -43,10 +43,10 @@ def add_arguments(self, parser):
 
     def decode_subject(self, subject):
         subject = decode_header(subject)
-        default_charset = 'ASCII'
+        default_charset = 'utf-8'
         # Convert the list of tuples containing the decoded string and encoding to
-        # the same encoding.
-        return u''.join([unicode(s[0], s[1] or default_charset) for s in subject])
+        # UTF-8
+        return u''.join([s[0].encode(default_charset, 'replace').decode(default_charset, 'replace') for s in subject])
```
jelly: This doesn't seem to be correct, tested what it should be:

```
-        return u''.join([s[0].encode(default_charset, 'replace').decode(default_charset, 'replace') for s in subject])
+        return ''.join([codecs.decode(s[0], s[1] or default_charset) for s in subject])
```

Requires an codecs import.

The test should be adjusted, main/tests/test_donor_import.py:
```
-        subject = str(Header(text, 'utf-8'))
+        subject = Header(text, 'utf-8')
```

Since s[0] are bytes.

README.md
twelveeighty: local_settings.py doesn't have existing values for EMAIL_HOST and EMAIL_PORT, so I changed the wording to say "add" vs. "change".

jelly: Yeah but note the escaping of _ which wasn't there ;-)

.gitignore
twelveeighty: Tbh, I'm new to 'coverage' as a tool. When it ran, it created a .coverage file that git wanted to add and commit, so I assumed it had to be ignored?

jelly: Yes, it can be ignored. But it has nothing to do with Python 3.

README.md
twelveeighty: Ah - Vim was showing it as a red parsing error without the escape, and it's escaped in other places, so I added them. I'm no markdown expert or purist, so I can take them out again, if you like?

main/management/commands/donor_import.py
twelveeighty: Agreed. I'll make your suggested changes and test. This one is tricky to test, so I'll have to rely on the unit tests.

jelly: I've tested my change with our data set and it works

twelveeighty: OK - I pushed another commit with your fix.

README.md
z3ntu: I think putting local_settings.py, email_host and email_port into a code block would probably be better

jelly:
Merged outside of Github

PR #102, opened 2018-04-27
Fix unit test warnings

twelveeighty:
Fix unit test warnings due to Paginators requiring specified ordering when rendering sitemaps. These were caused by calling order_by() without any parameters, which removes all ordering, even if specified at the entity level via Meta.


jelly:
Cherry picked in master, due to the faulty commits in this PR.

PR #125, opened 2018-06-24
Widen the search results

akstrfn:
Previously the search that had a space in it yields no results if the
match is not exact e.g. ``archlinux keyring`` yields no results even
though there is a package ``archlinux-keyring``. This patch modifies the
query to search for ``archlinux AND keyring``, both for package names
and package descriptions. This leads to more search results.

Hopefully this change makes sense :smile:

### DIFF
--- a/packages/views/search.py
+++ b/packages/views/search.py
@@ -1,4 +1,5 @@
 import json
+import operator
 
 from django import forms
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
@@ -98,8 +99,11 @@ def parse_form(form, packages):
 
     if form.cleaned_data['q']:
         query = form.cleaned_data['q']
-        q = Q(pkgname__icontains=query) | Q(pkgdesc__icontains=query)
-        packages = packages.filter(q)
+        q_pkgname = reduce(operator.__and__,
+                           (Q(pkgname__icontains=q) for q in query.split()))
+        q_pkgdesc = reduce(operator.__and__,
+                           (Q(pkgdesc__icontains=q) for q in query.split()))
+        packages = packages.filter(q_pkgname | q_pkgdesc)
 
     return packages
 


jelly:
I'm not sure yet, I haven't heard much complaints about not finding packages and I'm not sure about the performance impact for postgres.

akstrfn:
It bothered me enough to figure out django and implement it. I suppose it can be twice as slow?

akstrfn:
So after thinking this through it is very unlikely that it is twice as slow because that would be vary naive implementation approach which for sure is not used, but it would be good if you can test it? I'd test it but I have not idea how to do it.

For me this change is reasonable because when you search you usually don't know the name of the package and it is similar concept to ``find -name "*something*"`` vs `` find -name something``, where you almost always want the first one. New search tools (always?) "correct" this default eg. ``grep`` vs ``ag``, ``find`` vs ``fd`` etc.

Bluewind:
As long as you only use conjunctions (AND) the performance impact should be minimal. Any decent database will dismiss the row once one conjunct fails so adding additional conjuncts only incurs an impact for rows that already matched with the first keyword. Since you can only enter a single keyword right now, the performance won't get much worse. Obviously the additional keywords have to be validated against the found rows too, but those are far less than what is in the db so that will be fast. I'd say don't worry about that.

That said, I never actually checked this so feel free to dump the query (django surely has a query log or something, doesn't it?) that would run and run EXPLAIN on it. I doubt there will be weirdness though which is why I never bothered. Things look different if you were to use OR instead of AND obviously. Especially if you start searching with multiple disjunctions across multiple columns. You can get quite a lot of individual checks without the possibility of early termination in that case.

Also if you want a really fancy search that finds things if they are relatively similar, but not strictly equal, look at this: https://www.postgresql.org/docs/current/static/pgtrgm.html
There are some special index requirements and tradeoffs regarding the index (creation/update time and general index size I believe) so this isn't a no-brainer. Just throwing it out there since it's cool in general.

Issue #112, opened 2018-05-11
Dependencies on a virtual package contains an extra space


jelly:
For example the [package](https://www.archlinux.org/packages/community/x86_64/9base/) contains an extra space after 'bash' which shouldn't be there.

nehoffman:
I put in a pull request that should fix this. #128

PR #128, opened 2018-07-18
Fix For Issue #112

nehoffman:
Removes extra newline after the end span tag and before the li tag. This should fix issue 112 where there is an extra space after the last package. Because the change was so minor, I didn't test it out. It should work however since removing the newline from HTML in my web browser removed the extra space.


jelly:
Thanks, merged outside of the PR. I also reworded your long one line commit message and squashed them into one commit.

PR #150, opened 2018-10-30
Add retire_user command

jelly:


devel/management/commands/retire_user.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
+"""
+retire_user
+
+Retire an user by de-activing the account and moving the user to the retired
```
Bluewind: "Retire a user" instead of "an"

devel/management/commands/retire_user.py
```diff
@@ -0,0 +1,72 @@
...
+
+MAPPING = {
+        'Developers': 'Retired Developers',
+        'Trusted Users': 'Retired Trusted Users',
+        'Support Staff': 'Retired Support Staff',
+}
+
+
+class Command(BaseCommand):
+    help = "Retires a user, deactiving the user and moving the group membership to retired groups."
```
Bluewind: May replace the comma with a "by" so that it's more clear that the second part is performed by this function?

jelly: Fixed

devel/management/commands/retire_user.py
jelly: Fixed

PR #163, opened 2018-11-29
templates/public/about: remove unsupported section

SantiagoTorres:
The [unsupported] section blurb was outdated in both its existence and
the fact that there are almost 50k PKGBUILDs inside of it as of $DATE.


coderobe:
i'd leave the 9000 in :P

jelly:
It's not meaningful to specify the numbers since they change all the time when we delete or add packages.

jelly:
Although I like the 9000, I like the rephrasing of the sentence so I'll merge it.

PR #167, opened 2018-12-16
Switch to Django 2.x with Python 3.7

jelly:
Update archweb to Python 3.7 and Django 2.1.4.


archweb.wsgi
```diff
@@ -5,7 +5,7 @@ import site
 
 base_path = "/srv/http/archweb"
 
-site.addsitedir('/srv/http/archweb-env/lib/python2.7/site-packages')
+site.addsitedir('/srv/http/archweb-env/lib/python3.6/site-packages')
```
z3ntu: PR title says Python 3.7 but this says 3.6

jelly: Yeah this will not work.

PR #172, opened 2019-01-23
added docker section to /downloads

rowanruseler:


SantiagoTorres:
LGTM too, thanks for the suggestion @rowanruseler.

We will probably move it to the library soon :tm: but I'll update that reference as soon as that happens :)

PR #192, opened 2019-02-10
Attempt to quote pkgname to avoid SQL treat it as table.field

felixonmars:
This should fix the rdepend problem (qspectrumanalyzer not listed in "required by" of python-qt.py).


felixonmars:
Not really. Closing for now.

PR #214, opened 2019-04-14
[donotmerge] Unify command logging

### LINKED ISSUE #209: Refactor commands shared logging logic
All of our commands set up the logger in a similar way, this could be fixed by inheritance or figuring out if Django has some buildin logger logic for commands.

MrAnno:
Django does not currently have logging functionalities for commands:
https://code.djangoproject.com/ticket/21429

This PR introduces a preconfigured logger instance in `main.management.command.BaseCommand` that can be used for command logging.

The `--verbosity` option is taken into account:
  - 0: ERROR..WARNING
  - 1: ERROR..INFO
  - 2: ERROR..DEBUG
  - 3: unused

The logger instance is configurable, its name is `archweb.command`.
Errors and warnings are forwarded to stderr, everything else appears on stdout.

Colors are currently not supported.

Resolves #209

### DIFF
--- a/devel/management/commands/generate_keyring.py
+++ b/devel/management/commands/generate_keyring.py
@@ -7,20 +7,12 @@ Assemble a GPG keyring with all known developer keys.
 Usage: ./manage.py generate_keyring <keyserver> <keyring_path>
 """
 
-from django.core.management.base import BaseCommand, CommandError
+from django.core.management.base import CommandError
 
-import logging
 import subprocess
-import sys
 
+from main.management.command import BaseCommand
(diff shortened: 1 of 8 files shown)


MrAnno:
@jelly Could you have a look at this PR, please? I did not want to introduce a global logger variable, so I used inheritance. This way, I had to convert a few helper functions to command methods, hence the enormous diff.

If you like the concept, I'll change the remaining 6 commands as well.

jelly:
Hi, I think generally it looks nice, not sure if I would have done it this way. Sad to see that the django project  hasn't merged general logging.

PR #223, opened 2019-06-13
add security information page

SantiagoTorres:
After discussion with some of the security team, we decided that having more upfront information about disclosing vulnerability information with the security team would be useful.

This relates to [this issue](https://kanboard.archlinux.org/project/8/task/176) in the kanboard.

cc @Foxboron @jelly

### DIFF
--- a/public/views.py
+++ b/public/views.py
@@ -168,4 +168,20 @@ def keys_json(request):
     to_json = json.dumps(data, ensure_ascii=False)
     return HttpResponse(to_json, content_type='application/json')
 
+@cache_control(max_age=307)
+def security(request):
+
+    # FIXME: we may want to sort by firstname,lastname
+    developer_keys = DeveloperKey.objects.select_related(
+            'owner').filter(owner__isnull=False)
+
+    # FIXME this is probably not the way we want to do things for now.
+    SECURITY_TEAM = {'anthraxx', 'rgacogne', 'shibumi', 'jelle', 'sangy', 'Foxboron'}
+    developer_keys = [key for key in developer_keys if key.owner.userprofile.alias in SECURITY_TEAM]
+    context = {
+        'keys': developer_keys,
+    }
+    return render(request, 'public/security.html', context)
+
+
 # vim: set ts=4 sw=4 et:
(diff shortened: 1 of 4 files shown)


public/views.py
```diff
@@ -168,4 +168,20 @@ def keys_json(request):
...
 
+@cache_control(max_age=307)
+def security(request):
+
+    # FIXME: we may want to sort by firstname,lastname
+    developer_keys = DeveloperKey.objects.select_related(
+            'owner').filter(owner__isnull=False)
+
+    # FIXME this is probably not the way we want to do things for now.
+    SECURITY_TEAM = {'anthraxx', 'rgacogne', 'shibumi', 'jelle', 'sangy', 'Foxboron'}
```
jelly: Yeah, we should probably create a new group, like developers, trustedusers where these users belong :)

templates/base.html
```diff
@@ -24,7 +24,7 @@
                 <li id="anb-forums"><a href="https://bbs.archlinux.org/" title="Community forums">Forums</a></li>
                 <li id="anb-wiki"><a href="https://wiki.archlinux.org/" title="Community documentation">Wiki</a></li>
                 <li id="anb-bugs"><a href="https://bugs.archlinux.org/" title="Report and track bugs">Bugs</a></li>
-                <li id="anb-security"><a href="https://security.archlinux.org/" title="Arch Linux Security Tracker">Security</a></li>
+                <li id="anb-security"><a href="/security/" title="Arch Linux Security">Security</a></li>
```
jelly: Hmm I'm not sure if that's what we want. We still want people to find the tracker, but also find security contact information.

public/views.py
jelly: We have these fixtures here https://github.com/archlinux/archweb/blob/master/main/fixtures/groups.json which should become data migrations later.

Foxboron: Are all of these people on the email in the first place :x?

templates/base.html
Foxboron: I think  prefacing it with the team and then provide a link to the tracker is good enough in this case.

anthraxx: as said on the IRC im very much against this.
We can surely make the contact information page as "home" for the security tracker (if thats what you guys want), but the link should bring you to the security sub-space. that's where it belongs and info should be maintained.

SantiagoTorres:
Ok, it appears that there was some confusion. I'll migrate this information into security.archlinux.org.

Thanks for the reviews!

Issue #218, opened 2019-05-16
Send a notification email once a signoff package has reached the treshold


jelly:
When a package with signoff requirements has reached the required number of signoffs, send a notification email to its packager(s).

kyrias:
This can be closed now.

jelly:
Yes, except the small bug where it keeps sending emails!

PR #234, opened 2019-08-11
Send notification once package has enough signoffs

lrustand:
Implements issue #218

### DIFF
--- a/packages/views/signoff.py
+++ b/packages/views/signoff.py
@@ -4,10 +4,12 @@ from operator import attrgetter
 from django import forms
 from django.contrib.auth.decorators import permission_required
 from django.contrib.auth.models import User
+from django.core.mail import EmailMessage
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import transaction
 from django.http import HttpResponse, Http404
 from django.shortcuts import get_list_or_404, redirect, render
+from django.template import loader
 from django.utils.timezone import now
 from django.views.decorators.cache import never_cache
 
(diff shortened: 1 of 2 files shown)


packages/views/signoff.py
```diff
@@ -53,6 +55,36 @@ def signoff_package(request, name, repo, arch, revoke=False):
             return render(request, '403.html', status=403)
         signoff, created = Signoff.objects.get_or_create_from_package(
                 package, request.user)
+        signoffs = Signoff.objects.for_package(package).filter(
+                revoked__isnull=True)
+
+        signed_off_by = []
+        for s in signoffs:
+            signed_off_by.append(User.objects.get(id=s.user_id))
```
kyrias: So two things here, getting the `User` object is pointless because `s.user` would give it to you.

But more importantly, this isn't useful since it's only used for the template, so I would prefer if this variable was removed and `signoffs` was passed directly to the template.  Then the template just needs to do `s.user.username` instead.

packages/views/signoff.py
```diff
@@ -53,6 +55,36 @@ def signoff_package(request, name, repo, arch, revoke=False):
...
+
+        if  signoffs.count() >= spec.required:
+            packager = package.packager
+            if not packager:
+                toemail = [package.packager_str.split('<')[1].replace('>','')]
```
kyrias: This type of splitting is rather brittle, let's just not send out a notification if no matching ArchWeb user exists.

jelly: Agreed, let's drop this. Maybe it should go to a generic mail address, we can add a TODO for it for now.

templates/packages/approved.txt
```diff
@@ -0,0 +1,4 @@
+{% autoescape off %} The {{ pkg.repo.name }} package {{ pkg.pkgname }} {{ pkg.full_version }} has received the required number of signoffs.
+
+The package was signed off by the following users:{% for s in signoffs %}
+  {{ s.username }}{% endfor %}{% endautoescape %}
```
jelly: small nitpick maybe add a "-" for it to indicate it's a list.

packages/views/signoff.py
```diff
@@ -53,6 +55,36 @@ def signoff_package(request, name, repo, arch, revoke=False):
...
+            else:
+                toemail = [packager.email]
+            subject = '%s package [%s %s] approved' % \
+                    (package.repo.name, package.pkgname,
+                     package.full_version)
```
jelly: we should support f-strings now

lrustand: Thanks for the heads up! I was unaware of the f-strings functionality, I'm glad you told me about them.

jelly:
Thanks a lot for the PR!

PR #245, opened 2019-10-08
devel: add reposec management command

jelly:
Add a new command that scans pkg.tar.xz files with elf binaries in
/usr/bin/ and checks for security hardening issues. This adds a new
dashboard view which shows packages with these issues.


jelly:
Still a bit WIP

devel/management/commands/reposec.py
```diff
@@ -0,0 +1,237 @@
...
+
+    with Pool(processes=processes) as pool:
+        results = pool.map(partial(read_file, arch, repo), tasks)
+
+    results = [r for r in results if r]
+    results = [r for r in results if not r.is_pie or not r.is_relro or not r.is_canary]
+
+    with transaction.atomic():
+        PackageSecurity.objects.all().delete()
+        PackageSecurity.objects.bulk_create(results)
```
kyrias: Not sure if it makes a big difference, but in `reporead` we:
* Create ones that exist in the repo but not in the DB.
* Delete ones in the DB but not in the repo.
* Update ones that are in both.

I don't know how big the difference in performance is, but given how many packages we have, it wouldn't surprise me if it gets somewhat substantial.

If we want to do it in a similar way, I think something like this could work:

* Before reading the repo, get a list of all package names and versions for that repo from the DB, probably in a `{pkgname: pkgver}` dict.

* Iterate over each package in the pacman repo file:
  * `if pkgname not in pkgdict` -> add the package file path to the `to_add` list.
  * `if pkgname in pkgdict`
    * and the version is the same, pop the package name from the package dict.
    * and the version is different, leave the package in the dict so it gets deleted and also add the package file path to the `to_add` list. 

* Delete everything left in the package dict from the repo.
* Parse all package files in the `to_add` list and add them to the repo.

kyrias: (But if this turns out to be fairly quick on our repos as it is, feel free to ignore this.)

main/models.py
```diff
@@ -444,6 +444,17 @@ class Meta:
         db_table = 'package_files'
 
 
+class PackageSecurity(models.Model):
+    pkg = models.ForeignKey(Package, on_delete=models.CASCADE)
```
kyrias:
```
    pkg = models.OneToOneField(Package, on_delete=models.CASCADE)
```

will get you a single `Package.packagesecurity` attribute rather than having to deal with a list/set of one.

devel/reports.py
```diff
@@ -167,6 +167,20 @@ def non_existing_dependencies(packages):
...
+    filtered = []
+    packages_ids = packages.values_list('id',
+                                       flat=True).order_by().distinct()
+    packages = PackageSecurity.objects.filter(id__in=set(packages_ids))
+    for package in packages:
+        package.pkg.pie = 'PIE enabled' if package.is_pie else 'No PIE'
+        package.pkg.relro = 'Full RELRO' if package.is_relro else 'None'
+        package.pkg.canary = 'Canary found' if package.is_canary else 'No canary found'
+        package.pkg.fortify = 'Yes' if package.is_canary else 'No'
+        filtered.append(package.pkg)
```
kyrias: What about this?

```
    packages = packages.select_related('packagesecurity').order_by()
    for package in packages:
        pkgsec = package.packagesecurity
        package.pie = 'PIE enabled' if pkgsec.is_pie else 'No PIE'
        ...
        filtered.append(package)
```

kyrias:
Looks good overall!

devel/management/commands/reposec.py
```diff
@@ -0,0 +1,237 @@
...
+        for entry in pkg:
+            if not entry.isfile:
+                continue
+
+            # Retrieve pkgname
+            if entry.name == '.PKGINFO':
+                continue
+
+            if not entry.name.startswith('usr/bin/'):
+                continue
```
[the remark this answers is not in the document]
anthraxx: yeah, that would be nice as well. Also maybe have a better detection for ELF atrifacts, as stuff can also be put into /opt and possibly other locations

devel/management/commands/reposec.py
```diff
@@ -0,0 +1,237 @@
...
+
+    return pkgsec
+
+
+
+def read_repo(arch, source_dir, processes, options):
+    tasks = []
+
+    directory = os.path.join(source_dir, 'os', arch)
+    for filename in glob(os.path.join(directory, f'*{PKG_EXT}')):
```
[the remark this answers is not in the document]
anthraxx: definitively makes sense, however note that most likely when we switch to zstd you can have a limited time with multiple different extensions. we should maybe detect this in a more proof way

devel/reports.py
anthraxx: just a reminder that shared object files will not be PIE in the same attribute entry as binary applications are. however DSO objects are still fine when they are shared objects, as they are position independent when dynamically linked in by providing a random base per shared object.

devel/management/commands/reposec.py
```diff
@@ -0,0 +1,237 @@
...
+    pkgname = basename.rsplit('-', 3)[0]
+
+    with file_reader(filename) as pkg:
+        for entry in pkg:
+            if not entry.isfile:
+                continue
+
+            # Retrieve pkgname
+            if entry.name == '.PKGINFO':
+                continue
```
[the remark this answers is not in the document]
jelly: Yeah, I was wondering about using the PKGINFO instead of the rsplit.

devel/management/commands/reposec.py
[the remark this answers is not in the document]
jelly: This should be configurable yeah!

devel/management/commands/reposec.py
```diff
@@ -0,0 +1,237 @@
...
+            if not entry.name.startswith('usr/bin/'):
+                continue
+
+            fp = io.BytesIO(b''.join(entry.get_blocks()))
+            elf = Elf(fp)
+
+            if not elf.is_elf():
+                continue
+
+            pkg = Package.objects.get(pkgname=pkgname, repo=repo)
```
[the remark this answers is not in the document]
jelly: indeed, and I had some half code for it.

devel/management/commands/reposec.py
jelly: This sounds smart! Let me try to implement something like it.

jelly:
Current PR does not work on postgresql since it can't pickle the Package object due to being a memoryview in postgresql and binary in sqlite.

The PR also only checks the first PIE file and not the status of *all* PIE files in the package.

devel/reports.py
jelly: This is handled in the class Elf code.
```
          if "ET_DYN" in header['e_type']:
              if self.dynamic_tags("DT_DEBUG"):
                  return True
              return True # DSO is PIE
          return False```

devel/management/commands/reposec.py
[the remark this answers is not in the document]
jelly: now moved to BIN_PATHS, which I should rename to ELF_PATHS. Let's first make a working version and then extend coverage.

devel/management/commands/reposec.py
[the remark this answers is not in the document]
jelly: Added .zst, but since well this might change in the future due to @kyrias comment about re-doing the logic about reading packages / updates.

devel/management/commands/reposec.py
jelly: I have only tried [core] and I think we should do this efficient. Also currently it removes all PackageSecurity's while it could just remove for Arch.

jelly: So I'm not so sure how this would work. We don't have a list of parsed packages since this implementation only adds a PackageSecurity object for once a file has a "violation". This kinda sucks for [community] which is huge. So maybe this needs some re-designing or re-use the PackageFile object.

PR #248, opened 2019-10-16
Add information about official docker image

shibumi:
We have an official docker image now, let's mention that on our
`Download` page.

Signed-off-by: Christian Rebischke <chris@nullday.de>

### DIFF
--- a/templates/public/download.html
+++ b/templates/public/download.html
@@ -81,8 +81,8 @@
 
     <h3>Docker image</h3>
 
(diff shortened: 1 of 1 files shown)


shibumi:
@SantiagoTorres do we really need to mention that other image? I think it would be confusing.. do we do something with that account? We could either mirror the image there or add other images (for example an image with arch-linux security tracker). Could be interesting for our infra team.

PR #252, opened 2019-10-25
news: Add the MAILMAN_PASSWORD option and use it for sending authenticated emails

grazzolini:
Given that the whitelisting of the arch-announce email is not sufficient for validation, we are setting a poster password to make sure the list cannot be abused.

### DIFF
--- a/news/views.py
+++ b/news/views.py
@@ -1,5 +1,6 @@
 from django import forms
-from django.core.mail import send_mail
+from django.conf import settings
+from django.core.mail import EmailMessage
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404, redirect
 from django.template import loader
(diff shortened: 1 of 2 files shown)


news/views.py
```diff
@@ -43,12 +44,15 @@ def form_valid(self, form):
...
+                headers['Approved'] = settings.MAILMAN_PASSWORD
             template = loader.get_template('news/news_email_notification.txt')
-            send_mail('[arch-announce] %s' % newsitem.title,
-                      template.render(ctx),
-                      '"Arch Linux: Recent news updates: %s" <arch-announce@archlinux.org>' % newsitem.author.get_full_name(),
-                      ['arch-announce@archlinux.org'],
-                      fail_silently=True)
+            EmailMessage(subject='[arch-announce] %s' % newsitem.title,
+                      body=template.render(ctx),
+                      from_email='"Arch Linux: Recent news updates: %s" <arch-announce@archlinux.org>' % newsitem.author.get_full_name(),
```
jelly: An option is to use f-strings!

PR #257, opened 2019-11-06
Move logos

grazzolini:
Moved the sponsor logos and the donate link up the page, so they are more easily visible.


jelly:
LGTM

jelly:
As the sponsors didn't ask for it I'll drop this PR.

PR #263, opened 2019-12-14
Add Packages RSS feeds for "All Stable Repos" and "All Testing Repos"

donny85:
Hello,
on the bug tracker there is [FS#59329](https://bugs.archlinux.org/task/59329) task named `Added/removed packages feed for "all repos" should only include main repos`.

I added two new RSS feed groups for the packages updates and additions/removals.

One is for the stable packages from the  official repositories
and one is for the testing packages from the official repositories.
The corresponding links are generated on the /feeds/ page.


jelly:
Thanks! Merged!

PR #265, opened 2019-12-20
Adds responsive styles

ThatJohn:
This commit adds a few styles to improve rendering on mobile site and gets rid of a few conflicting  !important directives on a case-by-case basis..

### DIFF
--- a/sitestatic/archweb.css
+++ b/sitestatic/archweb.css
@@ -12,9 +12,11 @@
  * to help prevent other stylesheets from interfering.
  */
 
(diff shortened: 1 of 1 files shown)


sitestatic/archweb.css
```diff
@@ -205,6 +206,57 @@ h5 {
     margin-left: -300px;
 }
 
+
+
+/****** John *******/
```
jelly: John? Not sure what that comment means =)

Issue #261, opened 2019-11-25
Provide planet functionality


jelly:
planet.archlinux.org is Python2 and requires manual changing of configuration files by the devops team. Archweb's user model should be extended to add a new rss field for Arch staff to put their blog feed url.

TODO:
- [ ] Make the website field required when a user fills in a website_rss field
- [x] Figure out displaying of a summary with bleach or convert to plain text (for security reasons)
- [x] Create a new model for fellow Planet instances and make them manageable in the Django admin interface
- [x] Fix the /planet view for when there are no feed items or more then can be displayed
- [x] Review DB queries
- [x] Review caching consider caching ETags to reduce bandwidth https://pythonhosted.org/feedparser/http-etag.html#using-etags-to-reduce-bandwidth
- [x] Rename update_planet to something more sensible
- [ ] Add a short text to the planet view on how to add your own feed
- [x] Fix the CSS of the RSS icon in the Subscribe list
- [x] Remove FeedItems if they Feed limit is exceeded
- [x] Clean up feedItems when a user becomes inactive (or resigns)
- [x] Figure out Planet/News overlapping CSS
- [x] Author missing from RSS Feed
- [x] Author missing in the feed view http://127.0.0.1:8000/planet/

PR #267, opened 2020-01-05
Introduce planet functionality in archweb

jelly:
This change introduces a replacment for planet.archlinux.org which uses
a python 2 project to generate static html from multiple RSS feed
sources. For archweb a set of 'static' feeds can be created in the
django admin view for the Arch forums and other static feeds, archweb
users can add their own blog rss feed in their profile which will create
a Feed model.

When running the update_planet command, all Feed models are iterated
over and the rss feed is parsed. The latest FeedItem is queried matching
the current Feed model and every newer entry in the RSS feed is added as
new FeedItem. Since the body is also stored in the FeedItem there is a
limit to the amount of FeedItems per Feed configured in settings.py of
which the default is 25.

When a user is marked as inactive his Feed model and items are
removed automatically to avoid keeping stale data around.

Closes: #261


jelly:
Seems I forgot to publish the migrations, will do so later.

planet/management/commands/update_planet.py
```diff
@@ -0,0 +1,137 @@
+"""
+update_planet
+
+Imports all feeds for users who have filled in a valid website and website_rss,
+the amount of items imported is limited to the defined FEED_LIMT
```
jelly: Should probably be expanded..

planet/management/commands/update_planet.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if not feed.entries:
+            logger.info("error parsing feed: '%s', feed has no entries", url)
+            return feeds
+
+        for entry in feed.entries[:settings.RSS_FEED_LIMIT]:
+            item = self.parse_entry(entry, latest, url)
+
+            if not item:
+                continue
```
jelly: Since I move the code to handle the latest published feed item to parse_entry this keeps looping even if we already determined we need to "break"

planet/management/commands/update_planet.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if entry.get('description'):
+            summary = bleach.clean(entry.description, strip=True)
+            if len(summary) > 2000:
+                summary = truncatewords_html(summary, 100)
+            item.summary = summary
+
+        if entry.get('author'):
+            item.author = entry.get('author')
+        # fall back to archweb user
```
jelly: This is missing still!

sitestatic/archweb.css
```diff
@@ -482,6 +482,64 @@ table thead th.sorter-false {
         color: white;
     }
 
+/* home: news */
```
jelly: that looks wrong

sitestatic/archweb.css
```diff
@@ -482,6 +482,64 @@ table thead th.sorter-false {
...
+        border-bottom: 1px dotted #bbb;
+    }
+
+    #planet .timestamp {
+        float: right;
+        font-size: 0.85em;
+        margin: -1.8em 0.5em 0 0;
+    }
+
+/* home: arrowed headings */
```
jelly: Merge with home instead?

planet/management/commands/update_planet.py
```diff
@@ -0,0 +1,139 @@
...
+    level=logging.WARNING,
+    format=u'%(asctime)s -> %(levelname)s: %(message)s',
+    datefmt=u'%Y-%m-%d %H:%M:%S',
+    stream=sys.stderr)
+logger = logging.getLogger()
+
+
+class Command(BaseCommand):
+
+    def parse_feed(self, feedmodel):
```
kyrias: `feed` would feel nicer, since you're passing in an instance rather than the model class itself.

planet/management/commands/update_planet.py
```diff
@@ -0,0 +1,139 @@
...
+        url = feedmodel.website_rss
+
+        logger.debug("Import new feed items for '%s'", url)
+
+        try:
+            latest = FeedItem.objects.filter(feed=feedmodel).latest()
+        except FeedItem.DoesNotExist:
+            pass
+
+        etag = cache.get(url)
```
kyrias: Might want to namespace this by e.g. prepending `planet:etag:` to the URL.

planet/management/commands/update_planet.py
```diff
@@ -0,0 +1,139 @@
...
+    datefmt=u'%Y-%m-%d %H:%M:%S',
+    stream=sys.stderr)
+logger = logging.getLogger()
+
+
+class Command(BaseCommand):
+
+    def parse_feed(self, feedmodel):
+        latest = None
+        feeds = []
```
kyrias: This should be `feed_items` or just `items`.

planet/management/commands/update_planet.py
kyrias: A more Pythonic option would be to throw an `ItemOlderThanLatest` exception and handle this here.

planet/management/commands/update_planet.py
```diff
@@ -0,0 +1,139 @@
...
+            logger.info("error parsing feed: '%s', feed has no entries", url)
+            return feeds
+
+        for entry in feed.entries[:settings.RSS_FEED_LIMIT]:
+            item = self.parse_entry(entry, latest, url)
+
+            if not item:
+                continue
+
+            item.feed = feedmodel
```
kyrias: It would feel a bit cleaner to pass the feed to `parse_entry` and let it construct the `FeedItem` fully.

planet/management/commands/update_planet.py
```diff
@@ -0,0 +1,139 @@
...
+            feeds.append(item)
+
+        # TODO: only set cache when sucesfully changed?
```
kyrias: It would probably be better to move the `bulk_create` to just above this `cache.set`, yes.

planet/management/commands/update_planet.py
```diff
@@ -0,0 +1,139 @@
...
+        try:
+            latest = FeedItem.objects.filter(feed=feedmodel).latest()
+        except FeedItem.DoesNotExist:
+            pass
+
+        etag = cache.get(url)
+        feed = feedparser.parse(url, etag=etag)
+
+        if feed['status'] == 304:
+            logger.info("The feed '%s' has not changed since we last checked it", url)
```
kyrias: If we know that the feed hasn't changed, we should probably refresh the TTL on the cached ETag.

planet/management/commands/update_planet.py
```diff
@@ -0,0 +1,139 @@
...
+
+        if not published_parsed:
+            logger.error("feed: '%s' has no published or updated date", url)
+            return
+
+        published = datetime.fromtimestamp(time.mktime(published_parsed)).replace(tzinfo=utc)
+
+        # TODO: return False to stop iterating over feeds?
+        if latest and latest.publishdate >= published:
+            logger.debug("feed: '%s' has no new entries", url)
```
kyrias: This should say that it has no *more* new entries, since previous calls might have added some.

templates/planet/index.html
```diff
@@ -0,0 +1,64 @@
...
+<div id="intro" class="box">
+    <h2>Arch Planet</h2>
+
+    <p>Planet <strong>Arch Linux</strong> is a window into the world, work and
+    lives of Arch Linux hackers and developers.</p>
+</div>
+
+<div id="planet">
+    <h3>
+        <a href="{% url 'news-list' %}" title="Browse the planet archives">Latest Feeds</a>
```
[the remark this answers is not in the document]
jelly: I'm not sure yet, we now show 15 or 25 articles hmm maybe we should

planet/models.py
```diff
@@ -0,0 +1,55 @@
...
+    def __str__(self):
+        return self.title
+
+    class Meta:
+        db_table = 'feeditems'
+        verbose_name_plural = 'Feed Items'
+        get_latest_by = 'publishdate'
+        ordering = ('-publishdate',)
+
+class Planet(models.Model):
```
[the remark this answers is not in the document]
jelly: It is! Good point

planet/management/commands/update_planet.py
jelly: And return items...

planet/management/commands/update_planet.py
jelly: Good idea!

planet/management/commands/update_planet.py
```diff
@@ -0,0 +1,139 @@
...
+        item = FeedItem(title=entry.title, publishdate=published, url=entry.link)
+
+        if entry.get('description'):
+            summary = bleach.clean(entry.description, strip=True)
+            if len(summary) > FEEDITEM_SUMMARY_LIMIT:
+                summary = truncatewords_html(summary, 100)
+            item.summary = summary
+
+        if entry.get('author'):
+            item.author = entry.get('author')
```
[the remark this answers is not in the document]
jelly: Well it can be None.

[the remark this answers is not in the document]
kyrias: It'll be `None` either way though, so the conditional doesn't change the behaviour.  *shrg*

kyrias:
@jelly Haven't tried it yet, but looks good to me now!

kyrias:
(I've done some light testing now, and it seems to be working as expected.)

PR #268, opened 2020-01-06
netboot: add downgraded cipher banner

SantiagoTorres:
The banner is pretty self-explanatory. Inform users that the chainloaded
script is served on downgraded ciphers due to a limitation on the ssl
stack of ipxe. Also, add a configuration variable so that downstream
users of archweb can prevent the banner from displaying.

### DIFF
--- a/local_settings.py.example
+++ b/local_settings.py.example
@@ -54,4 +54,7 @@ MEDIA_URL = '/media/img/'
 ## Make this unique, and don't share it with anybody.
 SECRET_KEY = '00000000000000000000000000000000000000000000000'
 
+## Display a warning if serving netboot images on downgraded ciphers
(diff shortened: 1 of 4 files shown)


local_settings.py.example
```diff
@@ -54,4 +54,7 @@ MEDIA_URL = '/media/img/'
 ## Make this unique, and don't share it with anybody.
 SECRET_KEY = '00000000000000000000000000000000000000000000000'
 
+## Display a warning if serving netboot images on downgraded ciphers
+NETBOOT_SECURITY_BANNER=True
```
jelly: should be set as default false in settings I guess =)

jelly:
Merged locally with fixes.

PR #273, opened 2020-01-20
news: add note about news post procedure

jelly:


jelly:
https://www.mail-archive.com/arch-dev-public@archlinux.org/msg26234.html

PR #275, opened 2020-02-13
allow todo list name to wrap

eworm-de:
We have quite long todo list names, with attribute "nowrap" this looks
pretty strange on small screens.

### DIFF
--- a/templates/devel/index.html
+++ b/templates/devel/index.html
@@ -136,7 +136,7 @@
         <tbody>
             {% for todo in todos %}
             <tr>
-                <td><a href="{{ todo.get_absolute_url }}"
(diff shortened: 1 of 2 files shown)


jelly:
What kind of small screens? It looks quite ok but I have a 2048x1152 monitor at home.

eworm-de:
Opening the dashboard on mobile devices in portrait mode it looks like this.

Screenshot_20200213-213459

Not a deal breaker, but still ugly. Allowing to wrap would fix this. IMHO without downside.

jelly:
Merged and rebased via ea6a9d9357ff3c851ab80947a7af66c3c0c0d7c9

Issue #229, opened 2019-07-15
Switch IPy to Python3's ipaddress


jelly:
Use the native Python ipaddress module instead of the external IPy dependency for the RsyncModel.

PR #276, opened 2020-02-13
first attempt to solve #229 issue.

JAugusto42:
I remove the ipy, but not remove from the requirements, if this work, i remove from there.


jelly:
I think the issue with using ipaddress is that it does not handle ipv6 with a subnet mask.

jelly:
Closed https://github.com/archlinux/archweb/pull/293

PR #283, opened 2020-05-11
Add gitlab CI integration

jelly:
Signed-off-by: Jelle van der Waa <jelle@archlinux.org>

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,16 @@
+image: archlinux/base
+
+before_script:
+   - pacman -Syu --needed --noconfirm python-pip git
+   - pip install pylint coverage
+   - pip install -r requirements.txt
+   - python manage.py collectstatic --noinput
+
+lint:
+  script:
+     - pylint devel main mirrors news packages releng templates todolists visualize *.py
+
+# TODO: https://docs.gitlab.com/ee/ci/junit_test_reports.html
+test:
+  script:
+     - coverage run --rcfile .coveragerc manage.py test


jelly:
@svenstaro is this cool? :+1:

svenstaro:
Let's just merge it. We'll see whether it breaks. :)

.gitlab-ci.yml
```diff
@@ -0,0 +1,25 @@
+image: archlinux/base
```
FFY00, line 1: Should probably use `archlinux` over `archlinux/base`.

.gitlab-ci.yml
```diff
@@ -0,0 +1,25 @@
...
+    PIP_CACHE_DIR: "$CI_PROJECT_DIR/.cache/pip"
+
+cache:
+  paths:
+    - .cache/pip
+
+before_script:
+   - pacman -Syu --needed --noconfirm python-pip git
+   - pip install pylint coverage
+   - pip install -r requirements.txt
```
FFY00, line 15: Why pip since we have everything in the repos?

svenstaro: Well, not everything is in the repos. Also, we wouldn't want the project to break without knowing if something in Arch gets upgraded.

FFY00: Ah, sorry, looks like I missed some dependencies. But that's easily fixable. How are we deploying with missing dependencies?
If something updates and breaks archweb it would be good to know so that we can fix it, although I do see your point.

.gitlab-ci.yml
jelly: because?

.gitlab-ci.yml
jelly: For historical reasons we use pip, this might change in the future but indeed requires better / more up to date CI. Which is possible with gitlab, so it's probably a good idea to:
* package dependencies
* setup gitlab ci with packages and postgres/sqlite
* take it from there.

PR #314, opened 2020-10-16
Updated Readme.md for ease of user

hchaurasiya:


jelly:
This is a meaningless change, please don't do this.

PR #316, opened 2020-10-19
Add core.db.tar.zst support for reporead

jelly:
As Python does not support zstd compression yet, xtarfile a wrapper
around tarfile with zstd support is required.


anatol:
Woot! Thank you @jelly.

PR #321, opened 2020-11-26
Install `cssmin` using `poetry` compatible URL

abitrolly:
`poetry` doesn't support `requirements.txt` directly, but this works

    cat requirements.txt | xargs poetry add

Maybe consider switching to some more modern supported library.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
--e git+git://github.com/fredj/cssmin.git@master#egg=cssmin
+git+https://github.com/fredj/cssmin.git
 Django==3.1.3
 IPy==1.00
 Markdown==3.2.2


jelly:
Do you know a more modern css minifier which can be integrated in django?

abitrolly:
Plenty of choice there https://pypi.org/search/?q=css+minifier Never used one.

requirements.txt
```diff
@@ -1,4 +1,4 @@
--e git+git://github.com/fredj/cssmin.git@master#egg=cssmin
+git+https://github.com/fredj/cssmin.git
```
jelly, line 1: Reading the [documentation](https://pip.pypa.io/en/stable/reference/pip_install/#git) it seems for pip #egg=cssmin is required so can't it be kept? Why would poetry not understand something pip makes up?

FFY00: I think the point was removing the `-e` which poetry can't handle, not sure why the egg spec is being removed.

jelly: Just tested it and poetry can't handle that @master either which is rather sad

abitrolly: I would raise an issue with `pip` for deprecating ` #egg=cssmin` suffix that looks like a hack.

FFY00: It is absolutely not. And it is already deprecated as eggs themselves are deprecated.
https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support

abitrolly: Absolutely not what?

FFY00: It is absolutely not a hack.

abitrolly: Press :smile: if `#egg` is a hack, and :tada: if it is not. )

FFY00: It's fine to argue if this is or is not a hack, but please do not go open issues for features very deliberately documented. This is open source and I assure you all the pip maintainers, as is usual on open source projects, are busy, you will only be wasting their already limited time.

abitrolly: I am using `poetry`. So, what is the consensus Are you going to merge this? )

PR #323, opened 2020-12-05
Notify about recently updated packages

DragoonAethis:
This PR adds an extra green text line with "X minutes ago" to recently upgraded packages, next to the update date. This is to notify people that a given version might not yet be synchronized on various mirrors. Looks like this:

image

### DIFF
--- a/main/models.py
+++ b/main/models.py
@@ -1,5 +1,7 @@
 from itertools import groupby
 from pgpdump import BinaryData
+from pytz import utc
+from datetime import datetime
 
 from django.db import models
 from django.db.models import Q
@@ -132,6 +134,19 @@ class Package(models.Model):
         '''get a URL suitable for things like email including the domain'''
         domain = Site.objects.get_current().domain
         return f'{proto}://{domain}{self.get_absolute_url()}'
+
+    @property
+    def updated_mins_ago(self):
+        # Note: This uses the archweb metadata update time, not when the
+        # package was actually pushed to any repo. We don't have that
+        # easily available without adding more fields and scripts.
+        # See: https://github.com/archlinux/archweb/pull/323
+        now = datetime.utcnow().replace(tzinfo=utc)
+        return int((now - self.last_update).total_seconds()) // 60
+
+    @property
+    def is_recent(self):
+        return self.updated_mins_ago <= 90
 
     @property
     def signature(self):
--- a/sitestatic/archweb.css
+++ b/sitestatic/archweb.css
@@ -851,6 +851,12 @@ table.results {
         padding: 0.5em;
     }
 
+    #pkgdetails #pkginfo .recent {
+        color: green;
+        font-weight: bold;
+        text-decoration: underline dotted;
+    }
+
 /* pkgdetails: flag package */
 #flag-pkg-form label {
     width: 10em;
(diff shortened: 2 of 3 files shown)


FFY00 (approved):
Thanks, people keep flagging discord :disappointed:

DragoonAethis:
And whine about it on IRC :P Maybe it could be orange/red, or get highlighted on "flag package" link hover so that it's easier to notice right away?

main/models.py
```diff
@@ -133,6 +135,15 @@ def get_full_url(self, proto='https'):
...
         return f'{proto}://{domain}{self.get_absolute_url()}'
 
+    @property
+    def updated_mins_ago(self):
+        now = datetime.utcnow().replace(tzinfo=utc)
+        return (now - self.last_update).seconds // 60
+
+    @property
+    def is_recent(self):
+        return self.updated_mins_ago <= 90
```
jelly, line 145: Is this an educated guess or well informed?

DragoonAethis: I took the mirror sync delays from the [mirror status](https://www.archlinux.org/mirrors/status/), removed all mirrors with more than 6 hours of delay (around 100 of them), then averaged the rest (around 600) - that gives you 1h26min, so I just rounded it up to 90 minutes. This notifies about delays on about 400 mirrors.

Averaging mirrors below 10 hours gives 1h44mins (covers 420 mirrors), below 24 hours gives 3h06mins (covers 500 mirrors) and all mirrors average to about 4 hours (covers 550 mirrors). The longer this period is, the more annoying this notification ends up being for maintainers though, so I took the first average as a good starting point. If people keep complaining too much, it can be always bumped up.

DragoonAethis:
Updated to diff time on timedelta.total_seconds() and not timedelta.seconds, since that makes the message display every day (oops). Also added an underline to the "X mins ago" text so that it's a bit more obvious this has an alt text.

jelly:
Sorry for the delay, found a small bug:
```
2020-12-30 21:09 UTC (0.0 minutes ago)
--

DragoonAethis:
TIL floats with nothing after the decimal point will be printed as whole numbers by the template engine, but not if it's 0.0... I'll add a cast there.

Date of archweb import vs repo update import is a concern, but in this case the database doesn't contain any more concrete info about when a package was actually published in any repositories. One could check for the build date, but that breaks too eg with package rebuilds (where you might want to wait for tens of packages to be built before pushing). I'm not sure if there's any relatively clean way to do this without adding an extra repo_update DateTimeField, plus some extra logic for repo update scripts to set that in the archweb database. That sounds nastier than just being late with the archweb notification.

jelly:
> TIL floats with nothing after the decimal point will be printed as whole numbers by the template engine, but not if it's 0.0... I'll add a cast there.
> 
> Date of archweb import vs repo update import is a concern, but in this case the database doesn't contain any more concrete info about when a package was actually published in any repositories. One could check for the build date, but that breaks too eg with package rebuilds (where you might want to wait for tens of packages to be built before pushing). I'm not sure if there's any relatively clean way to do this without adding an extra repo_update DateTimeField, plus some extra logic for repo update scripts to set that in the archweb database. That sounds nastier than just being late with the archweb notification.

Yup, we'll take it as is and I'll try to remember that :+1:

DragoonAethis:
Fixed, added a comment for future reference.

DragoonAethis:
While we're around this topic, maybe I could also add something like a red bar on the flagging page to be extra sure? Something like this:

image

PR #325, opened 2020-12-28
alter donor_import.py to work as mda command

freswa:
While the approach to iterate on a maildir works fine, this needs
additional tooling. Taking an email on stdin allows to set the script as
mda command in e.g. fetchmail and process the mail on delivery without
storing it to a maildir.

### DIFF
--- a/main/management/commands/donor_import.py
+++ b/main/management/commands/donor_import.py
@@ -11,17 +11,19 @@ An example subject:
 
 Subject: Receipt [$25.00] By: John Doe [john.doe@archlinux.org]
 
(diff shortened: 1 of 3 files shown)


main/management/commands/donor_import.py
```diff
@@ -11,17 +11,19 @@
 
 Subject: Receipt [$25.00] By: John Doe [john.doe@archlinux.org]
 
-Usage: ./manage.py donor_import path/to/maildir/
+Usage: cat mail,md | ./manage.py donor_import
```
jelly, line 14: should it be mail.md or hmm mail.eml?

PR #328, opened 2021-01-11
Ignore case on packages exact match search

keysym:
As described in: https://wiki.archlinux.org/index.php/PKGBUILD#pkgname

"Package names should only consist of lowercase alphanumerics"

Since all packages will have a lowercase name, it seems logical to ignore case when searching.

### DIFF
--- a/packages/views/search.py
+++ b/packages/views/search.py
@@ -59,7 +59,7 @@ class PackageSearchForm(forms.Form):
             return []
         if 'q' not in self.cleaned_data:
             return []
-        return Package.objects.normal().filter(pkgname=self.cleaned_data['q'].lower())
+        return Package.objects.normal().filter(pkgname__iexact=self.cleaned_data['q'])
 
 
 def parse_form(form, packages):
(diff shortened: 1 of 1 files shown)


DragoonAethis:
I'm not aware of any packages with uppercase letters and the guidelines are against those, but makepkg doesn't prevent creating packages with such names - perhaps [matching with iexact](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#iexact) would be more universal, just in case?

lahwaacz:
There are packages with uppercase letters, e.g. [hunspell-en_US](https://archlinux.org/packages/extra/any/hunspell-en_US/). IIRC in the past even some libreoffice or firefox localizations had uppercase letters...

Hinigatsu:
@lahwaacz Nice catch!

I followed @DragoonAethis suggestion (check 227b06b4484e398982012eb89aeda0a00bb09074) so now the "ignore case" works even if the package has an uppercase letter.

jelly:
Merged via https://gitlab.archlinux.org/archlinux/github-pull-closer/-/merge_requests/1

PR #337, opened 2021-03-22
New website design

novachief101:
A new website design. Navbar is now blue and visible when scrolling, borders of boxes are removed, input elements have been changed a little and submit input buttons are blue. Also the navigation bar logo is changed to be white so it can be easily read on the blue navbar. RSS, Download and Magnet logo have been changed to look flatter. There are a few other subtle changes.


jelly:
I can not accept a change into Arch's home style as it is part of our identity. The Pull request is very hard to work with as all the changes are in one commit. Splitting them up is necessary for me to be able to review the changes.

novachief101:
I'm going to change it to make it more similar to the original website and split it into multiple commits then create another pull request. Hope it will be more similar to the original!

PR #338, opened 2021-03-28
Some changes

novachief101:
Changes to main website CSS: Sticky navigation bar (stays on top when scrolling), properly formatted archweb.css, light grey top border on footer, no outline on focused input and `code` elements are a little less yellow.

Changes to icons: Just less shadows, a little more change in the magnet icon.


jelly:
I am not sure why we would want a sticky navbar, I am not in favor of it and don't see the added value.

PR #357, opened 2021-06-27
Allow to download sig from "download from mirror"

### LINKED ISSUE #356: Make the signature available for "download from mirror"
The redirect logic for https://archlinux.org/packages/core/x86_64/filesystem/download/ does not follow signatures for download, as of now this is no issue  but in the future pacman might change this behaviour.

Change https://archlinux.org/packages/core/x86_64/filesystem/download/  to https://archlinux.org/packages/core/x86_64/filesystem/download

And allow redirection for https://archlinux.org/packages/core/x86_64/filesystem/download.sig

just1602:
Hello !

To fix this, I added a new url pattern that match `download.sig` and call the same `download` function with a new `sig` param to True.

It's been a while since the last time I did some django, so let me know if you'd like to see this fixed another way!

Fixes #356

### DIFF
--- a/packages/urls.py
+++ b/packages/urls.py
@@ -17,6 +17,7 @@ package_patterns = [
     url(r'^signoff/revoke/$', signoff.signoff_package, {'revoke': True}),
     url(r'^signoff/options/$', signoff.signoff_options),
     url(r'^download/$', display.download),
+    url(r'^download.sig/$', display.download, {'sig': True}),
     url(r'^sonames/$', display.sonames),
     url(r'^sonames/json/$', display.sonames_json),
 ]
--- a/packages/views/display.py
+++ b/packages/views/display.py
@@ -243,7 +243,7 @@ def files_json(request, name, repo, arch):
     return HttpResponse(to_json, content_type='application/json')
 
 
-def download(request, name, repo, arch):
+def download(request, name, repo, arch, sig=False):
     pkg = get_object_or_404(Package.objects.normal(),
                             pkgname=name, repo__name__iexact=repo, arch__name=arch)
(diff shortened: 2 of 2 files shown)


jelly (changes requested):
This does not really resolve the issue described in the ticket. The issue is that in the future pacman will append  ".sig" to the download url before redirect and as /download/.sig does not match this won't resolve the issue.

So we should rework the generated url to trim the trailing / and the same goes for the new download.sig url

just1602:
Maybe there's something I don't get, but currently, on my machine when I `curl -L localhost:8000/packages/core/x86_64/filesystem/download --output test` I get the archive of the filesystem package.

When I `curl -L localhost:8000/packages/core/x86_64/filesystem/download.sig` I get the sig file of the filesystem package.

Is this not what you want, what can I change to make it works as desired?

jelly:
> Maybe there's something I don't get, but currently, on my machine when I `curl -L localhost:8000/packages/core/x86_64/filesystem/download --output test` I get the archive of the filesystem package.
> 
> When I `curl -L localhost:8000/packages/core/x86_64/filesystem/download.sig` I get the sig file of the filesystem package.
> 
> Is this not what you want, what can I change to make it works as desired?

I see, but the packages page has download/, so fixed with:

```
-                <li><a href="download/" rel="nofollow" title="Download {{ pkg.pkgname }} from mirror">Download From Mirror</a>
+                <li><a href="download" rel="nofollow" title="Download {{ pkg.pkgname }} from mirror">Download From Mirror</a>>
```

jelly:
merged via cf628634bb2226a277df8ae32694f4777cabad47

PR #363, opened 2021-07-23
Update ipxe images

dvzrv:
This updates the iPXE images on our website using the iPXE images from the [ipxe](https://archlinux.org/packages/community/x86_64/ipxe/) package (until we have fully automated this away using https://gitlab.archlinux.org/archlinux/releng and will not require this manual workflow anymore).

The old images are affected by https://github.com/ipxe/ipxe/issues/407, which renders them unable to successfully complete a TLS handshake, due to our certificate chain being "too large".

This PR also updates the accompanying documentation and links (as the file names and the way we create the images have changed).


monkz:
Tested new IPXE images. They are functional in my tests via dhcp pxe boot via dnsmasq or even chainloaded from another iPXE instance.

templates/releng/netboot.html
```diff
@@ -46,18 +46,18 @@ <h3>Download</h3>
...
 <h4>UEFI</h4>
 
 <ul>
-<li><a href="{% static "netboot/ipxe.efi" %}">ipxe.efi</a> (<a href="{% static "netboot/ipxe.efi.sig" %}">PGP signature</a>) - x86_64 UEFI executable</li>
+<li><a href="{% static "netboot/ipxe-arch.efi" %}">ipxe-arch.efi</a> (<a href="{% static "netboot/ipxe-arch.efi.sig" %}">PGP signature</a>) - x86_64 UEFI executable</li>
 </ul>
 
 <p>You can build your own iPXE images compatible with netboot.
-For details, check out the <a href="https://aur.archlinux.org/packages/ipxe-netboot">ipxe-netboot AUR package</a>.</p>
+For details, check out the <a href="https://archlinux.org/packages/community/x86_64/ipxe/">ipxe package</a>.</p>
```
jelly, line 60: ideally we would not hardcode the url but obtain it from a pkg istance with  `{{ pkg.get_full_url }}`

So `netboot_info` has to be expanded and obtain the package and pass it to the template.

PR #371, opened 2021-09-08
Fix parsing of versions with epoch in rebuilderd

foutrelis:
The epoch component was not removed from the version and was ending up
as part of pkgver causing package lookups to fail. This meant that new
information about packages with epoch was not imported to archweb.


jelly (approved):

jelly:
Merged via 022530c5ea42ebeae48cb5dfe6ce59bf5eedb1c4

Issue #377, opened 2021-09-23
Add <Author> to planet feed


jelly:
The `author_name` can be set in our `feeds.py` PlanetFeed model.

https://docs.djangoproject.com/en/3.2/ref/contrib/syndication/

gregoiredx:
Hi @jelly, here is an implementation proposition for this: https://github.com/archlinux/archweb/pull/384.
Let me know what you think :slightly_smiling_face:

jelly:
Implemented in https://github.com/archlinux/archweb/pull/384

PR #384, opened 2021-11-18
Add author to planet feed

gregoiredx:
Add author name to planet feed (as suggested by https://github.com/archlinux/archweb/issues/377).

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -73,3 +73,12 @@ def developer_client(client, developer, userprofile, groups):
 def denylist(db):
     # TODO: create Denylist object
     call_command('loaddata', 'main/fixtures/denylist.json')
+
+
+@pytest.fixture(autouse=True)
+def use_dummy_cache_backend(settings):
+    settings.CACHES = {
+        "default": {
+            "BACKEND": "django.core.cache.backends.dummy.DummyCache",
+        }
+    }
--- a/feeds.py
+++ b/feeds.py
@@ -429,6 +429,9 @@ class PlanetFeed(Feed):
         return 'tag:%s,%s:%s' % (Site.objects.get_current().domain,
                                  date.strftime('%Y-%m-%d'), item.url)
 
+    def item_author_name(self, item):
+        return item.author
+
 
 def planet_last_modified(request, *args, **kwargs):
     try:
--- a/planet/tests/test_views.py
+++ b/planet/tests/test_views.py
@@ -1,6 +1,29 @@
+from datetime import datetime
+
+import feedparser
+import pytz
+
+from planet.models import FeedItem
+
+
 def test_feed(db, client):
     response = client.get('/feeds/planet/')
     assert response.status_code == 200
+    feed = feedparser.parse(response.content)
+    assert feed['feed']['title'] == 'Planet Arch Linux'
+
+
+def test_feed_item(db, client):
+    publishdate = datetime.now(pytz.utc)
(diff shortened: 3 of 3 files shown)


jelly:
Please have a look at the test failure

gregoiredx:
OK, the tests needed a `DATABASES` settings. Easiest way to allow that is to use the `local_settings.py.example` as the `local_settings.py` used by the CI, WDYT?

jelly:
> OK, the test needed a `DATABASES` settings. Easiest way to allow that is to use the `local_settings.py.example` as the `local_settings.py` used by the CI, WDYT? See [7d742a9](https://github.com/archlinux/archweb/commit/7d742a9c2b033f06c2f410b7b26921e8f9ba0d0d)

That can't be correct, it should never be required.

gregoiredx:
@jelly, thanks for your review. 
You're right, the failing test had nothing to do with the `DATABASES` setting.
What made the test suite pass when using a  `local_settings.py` in the CI was the `CACHES` config: https://github.com/archlinux/archweb/blob/099162395c01dc2df9289be1d9dd7a500becf4fe/local_settings.py.example#L34-L40

Because there is a response cache on `/feeds/planet/`: https://github.com/archlinux/archweb/blob/d108095aac9c6203e9e1a8bcbfe0e59624f64bf8/urls.py#L69
So the test `test_feed_item()` was working well when run alone but was getting an empty items list if called after another test already calling  `/feeds/planet/` in a context where no `FeedItem` existed.

So I introduced a dummy cache config in the `conftest.py`, exactly like proposed in [pytest-django documentation](https://pytest-django.readthedocs.io/en/latest/configuring_django.html#overriding-individual-settings).

I think it makes sense to deactivate caching in test settings to avoid side effects between tests, WDYT?

jelly:
> @jelly, thanks for your review. You're right, the failing test had nothing to do with the `DATABASES` setting. What made the test suite pass when using a `local_settings.py` in the CI was the `DummyCache` `CACHES` config:
> 
> https://github.com/archlinux/archweb/blob/099162395c01dc2df9289be1d9dd7a500becf4fe/local_settings.py.example#L34-L40
> 
> Because there is a response cache on `/feeds/planet/`:
> 
> https://github.com/archlinux/archweb/blob/d108095aac9c6203e9e1a8bcbfe0e59624f64bf8/urls.py#L69
> 
> 
> So the test `test_feed_item()` was working well when run alone but was getting an empty items list if called after another test already calling `/feeds/planet/` in a context where no `FeedItem` existed.
> So I introduced a dummy cache config in the `conftest.py`, exactly like proposed in [pytest-django documentation](https://pytest-django.readthedocs.io/en/latest/configuring_django.html#overriding-individual-settings).
> 
> I think it makes sense to deactivate caching in test settings to avoid side effects between tests, WDYT?

Nice catch, this rings a bell and I recall other tests do something with disabling / enabling the cache as well. I'll have a look over the weekend.

gregoiredx:
Yep, the following code could probably be removed if we apply the global test fixture in this PR:
https://github.com/archlinux/archweb/blob/cc799ff11127a9c665c83f5896bde42dc98287cb/mirrors/tests/test_mirrorstatus.py#L13
https://github.com/archlinux/archweb/blob/cc799ff11127a9c665c83f5896bde42dc98287cb/mirrors/tests/test_mirrorstatus.py#L43

jelly (approved):

jelly:
Thanks!

PR #388, opened 2021-12-12
Add theme-color meta tag

kusakata:
Add theme-color meta tag to header using archlinux-blue color (https://developer.mozilla.org/docs/Web/HTML/Element/meta/name/theme-color).


jelly:
Thanks!

jelly (approved):

PR #390, opened 2021-12-14
devel/views: Actually save first and last names when they change

kyrias:
Commit e4b211e8 (Allow users to set their own names, 2021-10-14) tried
to allow users to set their own names.  It correctly handles the latin
name field, but it doesn't end up actually saving changes to the first
and last name fields.

Signed-off-by: Johannes Löthberg <johannes@kyriasis.com>


jelly (approved):

jelly (approved):

jelly:
Thanks!

PR #391, opened 2021-12-23
Mutualize commands logger configs

gregoiredx:
This is a proposition for a single management command logger configuration as suggested by https://github.com/archlinux/archweb/issues/209.


gregoiredx:
Rereading the PR diff, I realized that I missed the fact that some logger use the WARNING level and others the INFO level.
I'm putting this to draft, I will get back on this latter.

jelly:
Looks great, thanks!

PR #403, opened 2022-02-06
Added a JSON search endpoint for groups. 

Torxed:
The result from the group search is all packages that belong to a specific group.
It could probably have been baked into the package search via `https://archlinux.org/packages/search/json/?group=gnome`.
But I would expect the endpoints to be identical across views so I added it under `groups/`.

### DIFF
--- a/packages/views/search.py
+++ b/packages/views/search.py
@@ -16,6 +16,32 @@ from ..models import PackageRelation
 from ..utils import attach_maintainers, PackageJSONEncoder
 
 
+class GroupSearchForm(forms.Form):
+    limit = forms.IntegerField(required=False, min_value=0)
+    page = forms.CharField(required=False)
+    repo = forms.MultipleChoiceField(required=False)
+    arch = forms.MultipleChoiceField(required=False)
+    name = forms.CharField(required=False)
+    desc = forms.CharField(required=False)
+    sort = forms.CharField(required=False, widget=forms.HiddenInput())
+
+    def __init__(self, *args, **kwargs):
+        show_staging = kwargs.pop('show_staging', False)
+        super(GroupSearchForm, self).__init__(*args, **kwargs)
+
+        repos = Repo.objects.all()
+
+        if not show_staging:
+            repos = repos.filter(staging=False)
+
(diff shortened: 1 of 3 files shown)


packages/views/search.py
```diff
@@ -151,6 +177,36 @@ def get_context_data(self, **kwargs):
...
+    }
+
+    if request.GET:
+        form = GroupSearchForm(data=request.GET,
+                                 show_staging=request.user.is_authenticated)
+
+        if form.is_valid():
+            form_limit = form.cleaned_data.get('limit', limit)
+            limit = min(limit, form_limit) if form_limit else limit
+            container['limit'] = limit
```
jelly, line 196: these two lines can be combined into one

jelly (changes requested):
It would be nice if there was a unit test for this new functionality.

packages/views/search.py
```diff
@@ -151,6 +177,36 @@ def get_context_data(self, **kwargs):
         context['search_form'] = self.form
         return context
 
+def group_search_json(request) -> HttpResponse:
+    limit = 250
+
+    container = {
+        'version': 2,
```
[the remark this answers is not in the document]
Torxed: Ah fair point, I assumed this was the overall API version or something similar. I'll correct it :)

packages/views/search.py
Torxed: Stole it from the package search hehe, I'll combine it :)

Torxed:
> It would be nice if there was a unit test for this new functionality.

That's a good idea, let me see if I can wrap something together real quick.

> First of, thanks for the PR!

Pleasure is all mine, major project that's been put together and I simply wanted to contribute a small part to it :) Keep it up!

jelly (approved):

jelly:
> The unit test is a bit rudimentary. I would love for it to check the content rather than just the exit code, but at least it's something to check if the endpoint is answering something and that it's 200 :)

Ok, I'll merge and deploy it. If you are up to expanding the tests that would be appreciated!

Torxed:
Thank you for the quick merge! This will help out archinstall quite a bit.
I'll learn how the tests are setup and what testdata there is and see what kind of improvements I can make!

PR #407, opened 2022-04-07
Fix pip install error

just1602:
The unauthenticated `git://` protocol has been disable on GitHub[^1] and we
need to use `https://` or `git@github.com:<user>/<repo>` to pull a repository.

[^1]: https://github.blog/2021-09-01-improving-git-protocol-security-github/


jelly:
Nice catch!

PR #412, opened 2022-06-28
Leave descriptions out of todolist overview tables

foutrelis:
With long descriptions the todolist overview table gets unwieldy. The
todo title should suffice for these views, and the description should
only be shown in the todo details page.


jelly:
No strong opinion here, so I'll let it through.

PR #422, opened 2022-08-14
todolists: restore ordering by status in list view

foutrelis:
After removing the description column, status has become the 7th column
instead of the 8th; fix the tablesorter call to use the correct number.

Fixes: 74e24ff4e6a9 ("Leave descriptions out of todolist overview tables")


jelly:
Maybe we shouldn't hardcode the amount of columns :)

jelly (approved):

PR #424, opened 2022-08-18
Fix SCM link of kde-unstable packages

felixonmars:
They are currently pointing to trunk, which isn't correct.


felixonmars:
Or they may be added as a field in `main/fixtures/repos.json` and referred here.

jelly:
Hmm this feels like a dirty workaround, I guess the Repo model needs to be extended with a a field which defines how the url is formatted. Maybe also in preparation for the move to Git?

jelly:
Let's close this as the Git migration will change things and needs more research/thought.

PR #436, opened 2022-12-03
archlinux.ipxe: do not perform GPG signature verification for the root file system image

nl6720:
Now that there have been three ISO releases with OpenSSL signed root file system image, the GPG-based verification option `verify=y` can be removed.

Doing this will allow releasing official ISOs without a GPG signed root file system image. The past few ISOs have been signed with an "Ephemeral Signing Key", so nothing of value will be lost.
This was done because it is complicated to run `mkarchiso` as root while passing it the user's GPG keys.

The envisioned future automated release process would also benefit by not tangling itself with GPG and only rely on code signing certificates.

Follow-up to commit 248908d4223b9ccf274fe19175b91d365a8501f8

/cc @dvzrv @Torxed


Torxed:
Sounds good! I didn't expect this when I pushed the original suggestion, but cool to see :)

dvzrv:
Thanks! This looks good!

PR #441, opened 2023-01-30
Add 'out-of-date' title to flagged packages in search page

Donlon:
This make it clear for those people who are curious about why the version text is in red.


jelly (approved):

jelly:
Thanks!

PR #446, opened 2023-02-15
main: show Gitlab links for package details pages

jelly:
Works for mysql++ / percona-server


jelly:
Now with populate signoff specification support.

main/utils.py
```diff
@@ -135,6 +136,16 @@ def groupby_preserve_order(iterable, keyfunc):
     return result
 
 
+def gitlab_project_name_to_path(name: str) -> str:
+    '''Convert a Gitlab project name to variant which the Gitlab encodes in
+    it's url / API for example mysql++ becomes mysqlplusplus.'''
```
polyzen, line 141:
```suggestion
    its URL / API for example mysql++ becomes mysqlplusplus.'''
```

jelly: Thanks, fixed.

PR #448, opened 2023-02-17
Automate mirror submission

Torxed:
Concept so far:
* [x] Public endpoint that doesn't require an arch account
* [x] Captcha to combat most spam
* [x] When submitted, the mirror is default `Active: False` and `Public: False`. These change to `True` when:
  * [ ] ~Upon *(automated?)* verification of `Completion: 100.0%`~
  * [x] After email to maintainers is sent and they manually change the two values after inspection.
* [x] Tier 1 requests gets automatically rejected *(with a nice email explaining why)* if:
  * [x] No valid Tier 2 in the database matching the same name and URL endpoints.
  * [x] Tier 2 is newer than `registered < MIN days`
  * [x] Tier 2 has a log containing multiple suspicious errors

rec

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
--e git+git://github.com/fredj/cssmin.git@master#egg=cssmin
+-e git+https://github.com/fredj/cssmin.git@master#egg=cssmin
 Django==4.1.6
 IPy==1.1
 Markdown==3.3.7


.gitignore
```diff
@@ -6,6 +6,7 @@ local_settings.py
 archweb.db
 archweb.db-*
 database.db
+/*.tar.gz
```
jelly, line 9: Please move this to a separate commit (Can even be a separate PR which we can already merge)

requirements.txt
```diff
@@ -1,4 +1,4 @@
--e git+git://github.com/fredj/cssmin.git@master#egg=cssmin
+cssmin==0.2.0
```
jelly, line 1: This is an unrelated change and points to a totally different version of cssmin.

templates/mirrors/mirror_submit.html
```diff
@@ -0,0 +1,34 @@
+{% extends "base.html" %}
+{% load package_extras %}
+{% block title %}Arch Linux - Pacman Mirrorlist Generator{% endblock %}
+
+{% block content %}
+<div id="mirrorlist-gen" class="box">
+
+    <h2>Mirror Request</h2>
+
+    <p>This page is meant as a replacement of the old way of registring as a Tier 1 or Tier 2 mirror. Previously this was done through <a href="https://bugs.archlinux.org/index.php?string=&project=1&search_name=&type%5B%5D=&sev%5B%5D=&pri%5B%5D=&due%5B%5D=&reported%5B%5D=&cat%5B%5D=43&status%5B%5D=open&percent%5B%5D=&opened=&dev=&closed=&duedatefrom=&duedateto=&changedfrom=&changedto=&openedfrom=&openedto=&closedfrom=&closedto=&do=index">https://bugs.archlinux.org</a> and would require manual intervention from the mirror maintainer(s). This process is now semi-automated and various checks against your mirror will be performed when submitting via the below form.</p>
```
jelly, line 10: This url can be way shorter like:

https://bugs.archlinux.org/index.php?string=&project=1&search_name=&type%5B%5D=&sev%5B%5D=&pri%5B%5D=&due%5B%5D=&reported%5B%5D=&cat%5B%5D=43&status%5B%5D=open&percent%5B%5D=&do=index

mirrors/views/mirrorlist.py
```diff
@@ -127,4 +242,92 @@ def find_mirrors_simple(request, protocol):
     proto = get_object_or_404(MirrorProtocol, protocol=protocol)
     return find_mirrors(request, protocols=[proto])
 
+
+def mail_mirror_admins(data):
+    template = loader.get_template('mirrors/new_mirror_mail_template.txt')
+    for mirror_maintainer in ['anton@hvornum.se', 'pitastrudl@gmail.com']:
```
jelly, line 248: This can be looked up (queryied from the users db) instead of hardcoding.

templates/mirrors/mirror_submit.html
```diff
@@ -0,0 +1,34 @@
...
+
+    <h3>Available mirrors</h3>
+
+    <p>Below are direct links to the two different tiers you will need to be acquainted with.<p>
+
+    <ul>
+        <li><a href="/mirrors/tier/1/">Tier 1 mirrors</a></li>
+        <li><a href="/mirrors/tier/2/">Tier 2 mirrors</a></li>
```
jelly, line 18: ideally this uses the [url template tag](https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#url)

templates/mirrors/mirror_submit.html
```diff
@@ -0,0 +1,34 @@
...
+    </ul>
+
+    <h3>Mirror information</h3>
+
+    <p>Before you can submit a <b>Tier 1</b> request the mirror in question must first be a registered <b>Tier 2</b> for a certain amount of time with proven reliablity. Once the submitted information is verified the mirror will be visible under the appropriate tier list above. This process usually takes 5 minutes.</p>
```
jelly, line 28: The way I would interpert this is that my mirror will be accepted within 5 minutes after submission that sounds a bit optimisitic :)

jelly (commented):
Cool! I've made some minor comments in the PR. Would be nice to also add a test later, happy to help if you have some questions.

.gitignore
Torxed: Sure :)

requirements.txt
Torxed: Can revert this one, but `pip install -r requirements.txt` wouldn't run with it. Couldn't be bothered investigating why so matched the name against pypi and went with it :)

templates/mirrors/mirror_submit.html
Torxed: Thanks, just copied the URL from a default search of "mirror" category.

mirrors/views/mirrorlist.py
Torxed: Yepp, will get to it once I figure out where they're stored :)

templates/mirrors/mirror_submit.html
Torxed: Good idea, I'll learn how those work and see if I can get it working.

templates/mirrors/mirror_submit.html
Torxed: We can tweak it. But there's not much information in general what's the guiding values regarding any mirror related tasks.
5 min was a ballpark figure after having administrated the mirror list for a few days now. It usually takes less for the mirror "Completion" rate to be registered.
The idea was to automate the `Inactive` -> `Active` status once the first sync/check is performed.

For now the process will remain manual, and yea it will take quite a bit longer than 5 minutes hehe.

Just want to remind that this is a draft, and an attempt to convert the [manual process](https://wiki.archlinux.org/title/DeveloperWiki:NewMirrors) into at the very least semi-automated one ^^

Torxed:
> Cool! I've made some minor comments in the PR. Would be nice to also add a test later, happy to help if you have some questions.

Thanks for the feedback, me and the misses will get onto fixing the last few things before I mark this as ready instead of WIP.

The whole Django thing is still new to me and I have no idea where things are stored in the db :see_no_evil:

.gitignore
Torxed: I just noticed there was already a `*.tar.gz` that doesn't trigger on tar files in the root directory. I'll tweak that line instead of adding another one.

Torxed: https://github.com/archlinux/archweb/pull/449

templates/mirrors/mirror_submit.html
Torxed: Not sure how to work around the fact that the tier url's use `re_path(r'^tier/(?P<tier>\d+)/$', mirrors, name='mirror-list-tier'),` with one common name *(I assume that's how those links work after reading up)*.

Torxed:
Oh, and I will run a code linter once I feel ready. Just can't stand the syntax while developing :)

mirrors/views/mirrorlist.py
Torxed: Fixed :) Easier than I imagined.

Torxed:
Mail function looks good:
screenshot

Aside from automatic check to make the mirror public after certain criterias, I think the PR is mostly ready.

Torxed:
@jelly I consider the PR ready, as I might need help with writing tests for this.
But as mentioned, I can't figure out why the pip+git breaks (even in the runner):
```
Obtaining cssmin from git+git://github.com/fredj/cssmin.git@master#egg=cssmin (from -r requirements.txt (line 1))
  Cloning git://github.com/fredj/cssmin.git (to revision master) to ./src/cssmin
  Running command git clone --filter=blob:none --quiet git://github.com/fredj/cssmin.git /home/runner/work/archweb/archweb/src/cssmin
  fatal: unable to connect to github.com:
  github.com[0: 140.82.112.3]: errno=Connection timed out

  error: subprocess-exited-with-error
  
  × git clone --filter=blob:none --quiet git://github.com/fredj/cssmin.git /home/runner/work/archweb/archweb/src/cssmin did not run successfully.
  │ exit code: 128
  ╰─> See above for output.
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× git clone --filter=blob:none --quiet git://github.com/fredj/cssmin.git /home/runner/work/archweb/archweb/src/cssmin did not run successfully.
│ exit code: 128
╰─> See above for output.
```

mirrors/tests/test_mirror_submission.py
```diff
@@ -0,0 +1,3 @@
+def test_mirror_registration(client, mirrorurl):
+    response = client.get('/mirrorlist/submit/?name=test3&tier=2&upstream=1&admin_email=anton%40hvornum.se&alternate_email=&isos=on&rsync_user=&rsync_password=&notes=&active=True&public=True&url1-url=rsync%3A%2F%2Ftest3.com%2Farchlinux&url1-country=SE&url1-bandwidth=1234&url1-active=on&url2-url=&url2-country=&url2-bandwidth=&url2-active=on&url3-url=&url3-country=&url3-bandwidth=&url3-active=on&ip=&captcha_0=d5a017cc3851fb59898167f666759c99b42afd52&captcha_1=tdof')
```
Torxed, line 2: It's a crude test, but will at least test that the endpoint exists and that it accepts the request data *(captcha will fail)*

Torxed:
A thought that hit me while administrating the manual process was that adding a bunch of URL's to the same mirror might be desirable.

That currently won't work with this approach, but very well could in v2.0 of this endpoint.
I feel it's a bit of an undertaking adding support for 3+ URL's, especially dynamically added URL's.
So if no one objects, I'd prefer if this goes in before I start working on v2.0 of this.

jelly:
This PR still has CI issues.

requirements.txt
```diff
@@ -5,6 +5,7 @@ Markdown==3.3.7
 bencode.py==4.0.0
 django-countries==7.3.2
 django-extensions==3.1.3
+django-simple-captcha==0.5.17
```
jelly, line 8: Not sure if I like the captch dependency, we don't have it for flagging packages. Do you think it's really required?

jelly (changes requested):
Please / rebase cleanup the comments and I'll take another look

AUTHORS
```diff
@@ -37,3 +37,5 @@ Thayer Williams <thayer@archlinux.org>
 Thomas Bächler <thomas@archlinux.org>
 Tom Willemsen <tom.willemsen@archlinux.us>
 Tyler Dence <tyzoid.d@gmail.com>
+Anton Hvornum <anton@hvornum.se>
+Nina Nick <5041877+ninchester@users.noreply.github.com>
```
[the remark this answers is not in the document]
Torxed: Separate commit to the same PR, or separate PR with a separate commit?

mirrors/views/mirrorlist.py
```diff
@@ -234,41 +233,64 @@ def find_mirrors_simple(request, protocol):
     proto = get_object_or_404(MirrorProtocol, protocol=protocol)
     return find_mirrors(request, protocols=[proto])
 
+
```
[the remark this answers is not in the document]
Torxed: I'm sorry, but [PEP8#Blank-Lines](https://peps.python.org/pep-0008/#blank-lines) dictates two blank lines between function definitions. The rest of the code is also spaced out this way.

For reference: https://github.com/archlinux/archweb/blob/3904bac6e14b184fc438c9fe91c92a4c53dcc7d4/mirrors/views/mirrorlist.py#L191-L194

requirements.txt
Torxed: Well this is what people said I should put in when I was asking in the arch channels, I don't mind dropping it but to reduce spam this is what was suggested.

To quote: *"Just make a captcha thing"*

templates/mirrors/mirror_submit.html
```diff
@@ -0,0 +1,34 @@
...
+
+    <form id="list-generator" method="get">
```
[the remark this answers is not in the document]
Torxed: And just for reference, I took the method from this form:
* https://github.com/archlinux/archweb/blob/ad3b1828b4b52af1aff9a9e7fa29518327197384/templates/mirrors/mirrorlist_generate.html#L38-L41

mirrors/tests/test_mirror_submission.py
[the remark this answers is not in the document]
Torxed: Same as below:
 * https://github.com/archlinux/archweb/blob/ad3b1828b4b52af1aff9a9e7fa29518327197384/templates/mirrors/mirrorlist_generate.html#L38-L41

requirements.txt
```diff
@@ -1,5 +1,5 @@
 cssmin==0.2.0
-Django==4.0.1
+Django==4.1.6
```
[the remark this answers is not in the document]
Torxed: Sure, but saw an opportunity to update old stuff seeing as they have a few vulns in 4.0 that should be fixed in 4.1. But I'll revert it and someone can update later.

.gitignore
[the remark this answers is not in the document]
Torxed: Sure it's "unrelated", but the fact remains that whenever you work from the project folder and follow some instructions about downloading a database or something similar you run the risk of pushing it globally.

If the project does not need `.tar.gz` files to be built or executed, there's really no risk of adding it to the ignore list. And I thought while I'm at it, I'll add it.

requirements.txt
```diff
@@ -1,4 +1,4 @@
--e git+git://github.com/fredj/cssmin.git@master#egg=cssmin
+-e git+https://github.com/fredj/cssmin.git@master#egg=cssmin
```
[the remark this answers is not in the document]
Torxed: Changing.

PR #449, opened 2023-02-18
Tweaked git ignore to ignore all tar.gz and __pycache__

Torxed:


.gitignore
```diff
@@ -11,7 +11,9 @@ collected_static/
 testing/
 env/
 htmlcov
-*.tar.gz
+**/**.tar.gz
+/.idea*
```
jelly, line 15: I'll let this slip, but ought to really be a user gitignore instead of an per project ignore. Like I have:

git config --global core.excludesfile ~/.globalgitignore

https://github.com/jelly/Dotfiles/blob/master/.globalgitignore

jelly (approved):

.gitignore
Torxed: Ideally yes, but it's also a sane idea to have it there so the repo doesn't end up with garbage (it was close this time, as we set up two new computers to push for this project) hehe.

Thanks for merging :)

PR #455, opened 2023-03-13
enhancement: add transparent background to sponsor images

christian-heusel:
Currently the images sometimes were not using transparent background or were a little pixelated, which this PR tries to fix.

Old:
image

New:
image


lcarilla (approved):
lgtm :) 
really looks nicer this way

templates/public/donate.html
```diff
@@ -50,7 +50,7 @@ <h3>Commercial sponsors and contributions</h3>
     of git commits and software packages, GnuPG keys, and SSH accounts.</p>
 
     <a href="https://nitrokey.com/" title="Nitrokey">
-    <img src="{% static "nitrokey_logo.png" %}"
+    <img width="240" src="{% static "nitrokey_logo.png" %}"
```
jelly, line 53: Why not resize the original image back to 420px?

jelly: This logo is only shown here, it was 240px width, so we can just retain that:

```
magick convert -resize 240 nitrokey_logo.png nitrokey_logo_res.png
```

jelly (commented):
Please also oxipng all the images so they are about 10-20% smaller.

jelly (approved):
oxipng'd the icons, resized the nitrokey logo. All good now.

jelly:
Thanks!

PR #459, opened 2023-05-19
Add link to GitLab forge site

alerque:
The Arch Linux GitLab instance has been public for some time and
especially with package migration to Git is increasingly relevant. Even
without packages lots of Arch projects are tracked there and eventually
it will eclipse Flyspray in relevance for collecting bug reports. Hence
placing it near and before the "Bugs" link to Flyspray.


jelly:
I am not a fan of adding a lot of items to the top bar, we also can't only update archweb we need to update every the wiki, bbs, bugs, etc. So this will have to be postponed until we can drop flyspray.

PR #464, opened 2023-05-23
packages: supply a pkgbase/maintainer mapping

jelly:
For our Gitlab bot to assign members automatically to packaging repository a mapping is needed of pkgbase to maintainers.

### DIFF
--- a/packages/urls.py
+++ b/packages/urls.py
@@ -28,6 +28,7 @@ urlpatterns = [
     path('signoffs/json/', signoff.signoffs_json, name='package-signoffs-json'),
     path('update/', views.update),
     path('sonames', views.sonames),
+    path('pkgbase-maintainer', views.pkgbase_mapping),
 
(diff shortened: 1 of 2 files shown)


packages/urls.py
```diff
@@ -28,6 +28,7 @@
     path('signoffs/json/', signoff.signoffs_json, name='package-signoffs-json'),
     path('update/', views.update),
     path('sonames', views.sonames),
+    path('pkgbase-maintainer', views.pkgbase_mapping),
```
jelly, line 31: I really dislike this endpoint name, any better ideas?

anthraxx: I can totally relate why you dislike this endpoint name.

It's a hard one... I have no silver bullet... but how about simply `maintainers`? We already have like `flag` and `download` etc, thats all expected to be related to packages. so `maintainers` in that sense would simply be the endpoint for getting maintainers of packages?

PR #469, opened 2023-07-03
devel: rest api token auth

jelly:
Implement a simple token authentication header for various "restish" endpoints we might want for adoption/disowning of packages.

### DIFF
--- a/devel/models.py
+++ b/devel/models.py
@@ -7,11 +7,53 @@ from django.db.models.signals import pre_save, post_save
 from django.contrib.auth.models import User, Group
 from django_countries.fields import CountryField
 from django.core.validators import MinValueValidator, MaxValueValidator
+from django.core.exceptions import ImproperlyConfigured
+from django.utils.deprecation import MiddlewareMixin
 
 from .fields import PGPKeyField
 from main.utils import make_choice, set_created_field
 
 from planet.models import Feed
+
+
+class AuthTokenBackend(object):
+    def authenticate(self, request, username=None, password=None):
+        if 'X-Archweb-Token' in request.headers:
+            try:
+                profile = UserProfile.objects.get(api_token=request.headers['X-Archweb-Token'])
+            except UserProfile.DoesNotExist:
+                return None
+
+            print(profile)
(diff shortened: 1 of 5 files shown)


raffomania:
Hey, just a reminder, if you don't find the time to continue this, I'd like to help out :)

devel/models.py
```diff
@@ -7,13 +7,55 @@
...
 
 from .fields import PGPKeyField
 from main.utils import make_choice, set_created_field
 
 from planet.models import Feed
 
 
+class AuthTokenBackend(object):
+    def authenticate(self, request, username=None, password=None):
+        if 'X-Archweb-Token' in request.headers:
```
anthraxx, line 21: Wouldn't it make sense to use `Authorization: Bearer` as header, this is a quite widely used way for such API and would also allow a lot of HTTP clients to use implemented helper methods for authenticate a request with a token. Otherwise you'd need to pass around custom header strings like `X-Archweb-Token`.
What are your thoughts about this?

templates/devel/profile.html
```diff
@@ -23,5 +23,20 @@ <h2>Developer Profile</h2>
         <p><label></label> <input title="Save changes" type="submit" value="Save" /></p>
     </form>
 
+    <form id="api-profile-form" enctype="multipart/form-data" method="post" action="">{% csrf_token %}
+        <h3>API token</h3>
+        <p>Token for completing todolist items with for example, rebuild-todo</p>
+        {% if profile.api_token is None %}
```
anthraxx, line 29: I assume it would be a good pattern to be able to revoke a token, or to phrase it differently: Generate a new token and invalidate the existing one.
I'd probably propose something like only showing the token once when the generate button is clicked and otherwise show an invalidate/regenerate button instead. We should hide the token after generating it.

anthraxx (commented):
Awesome to see progress on this, this will be a very useful feature. Thank you for working on this.

devel/models.py
jelly: Sounds fine to me, it shouldn't conflict with things I suppose.

PR #483, opened 2023-11-13
templates/public/download.html: update checksum and signature list and verification

nl6720:
Link to *sum.txt files on archlinux.org, just like with the GPG signatures, as the mirrors cannot be trusted.

Add the GnuPG verification command from https://wiki.archlinux.org/title/Installation_guide#Verify_signature. It would be better if we do not duplicate it in both archweb and the wiki.


nl6720:
Partial screenshot:

Screenshot 2023-11-13 at 17-05-39 Arch Linux - Downloads

jelly (approved):

PR #490, opened 2024-03-03
Update flaghelp.html with link to GitLab

serebit:
Now that the bugtracker has been moved to GitLab, the flaghelp text should match.

### DIFF
--- a/templates/packages/flaghelp.html
+++ b/templates/packages/flaghelp.html
@@ -24,7 +24,8 @@
         with your additional text.</p>
 
         <p><strong>Note:</strong> Please do <em>not</em> use this facility if the
-        package is broken! Use the <a target="_blank" href="https://bugs.archlinux.org/"
-            title="Arch Linux Bugtracker">bugtracker</a> instead.</p>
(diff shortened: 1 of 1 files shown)


templates/packages/flaghelp.html
```diff
@@ -24,7 +24,8 @@ <h3>Flagging Packages</h3>
         with your additional text.</p>
 
         <p><strong>Note:</strong> Please do <em>not</em> use this facility if the
-        package is broken! Use the <a target="_blank" href="https://bugs.archlinux.org/"
-            title="Arch Linux Bugtracker">bugtracker</a> instead.</p>
+        package is broken! File an issue on the package's <a target="_blank" href="https://gitlab.archlinux.org/archlinux/packaging/packages"
```
christian-heusel, line 27: I think this should also point the user to the correct GitLab repo directly, see https://github.com/archlinux/archweb/commit/fca3ff5166d227ed4474ea6804fa748bc572371a for the related changes.

lahwaacz: There is not enough context to make the link more specific. This template is rendered at https://archlinux.org/packages/flaghelp/ (i.e. it is not bound to a specific package).

jelly (approved):
Thanks

PR #492, opened 2024-03-06
static: use a bigger favicon

felixonmars:
Our main website currently has a very small favicon that looks blurred on HiDPI browsers.

This png version came from our GitLab instance which looks much better.

Comparison:
image


jelly (approved):
Sure!

PR #494, opened 2024-03-10
Draft: Integrate archlinux-common-style

lahwaacz:
This depends on the following currently open merge requests:

- https://gitlab.archlinux.org/archlinux/archlinux-common-style/-/merge_requests/9
- https://gitlab.archlinux.org/archlinux/archlinux-common-style/-/merge_requests/10

When they are merged, the URL of the `archlinux_common_style` git submodule must be updated to upstream.


jelly:
Sorry for checking this out so late, this basically switches it to use the `css` files from archlinux-common-styles no? Afaik Django does not handle sass?

lahwaacz:
Basically, yes, but more - with those merge requests for archlinux-common-style, the following would be taken from the external repo:

- the navbar's HTML (as a Django template)
- CSS for the navbar
- Arch logo and icons

 As for SASS, I have not managed to solve this part nicely yet. I found several projects with different approaches to the actual SASS backend, Django integration, and maintenance status. But all of them seem to focus on "live SASS builds" for development, which we specifically do not need for integrating an external project. We need to chat about this sometime :smile:

PR #498, opened 2024-04-19
releng/models.py: update bootstrap tarball extension

nl6720:
archiso v77 changed the bootstrap tarball compression from gzip to zstd. See https://gitlab.archlinux.org/archlinux/archiso/-/merge_requests/373

Related to https://gitlab.archlinux.org/archlinux/archiso/-/issues/130

/cc @dvzrv

### DIFF
--- a/releng/models.py
+++ b/releng/models.py
@@ -47,7 +47,7 @@ class Release(models.Model):
         return "iso/%s/archlinux-%s-x86_64.iso" % (self.version, self.version)
 
     def tarball_url(self):
-        return "iso/%s/archlinux-bootstrap-%s-x86_64.tar.gz" % (self.version, self.version)
+        return "iso/%s/archlinux-bootstrap-%s-x86_64.tar.zst" % (self.version, self.version)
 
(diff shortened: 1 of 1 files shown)


releng/models.py
```diff
@@ -47,7 +47,7 @@ def iso_url(self):
         return "iso/%s/archlinux-%s-x86_64.iso" % (self.version, self.version)
 
     def tarball_url(self):
-        return "iso/%s/archlinux-bootstrap-%s-x86_64.tar.gz" % (self.version, self.version)
+        return "iso/%s/archlinux-bootstrap-%s-x86_64.tar.zst" % (self.version, self.version)
```
dvzrv, line 50: Hm, wouldn't this also change for all currently available (and those not yet being zstd compressed)?

nl6720: `tarball_url` is used in only one place.
https://github.com/archlinux/archweb/blob/8679a0c0976a92c76fa307bba8ba0a1d2ea320df/templates/public/download.html#L117

PR #500, opened 2024-04-24
templates/public/download.html: fix mismatched closing tag

nl6720:
Reported in https://bbs.archlinux.org/viewtopic.php?id=295151

Fixes: b018f09fcada5a3b71d9b91bacfd1ddee9af9952 ("templates/public/download.html: update checksum and signature list and verification")


jelly (approved):
Woops!

PR #513, opened 2024-07-20
Optimise package template / caching

jelly:


templates/packages/package_details.html
```diff
@@ -213,6 +214,7 @@ <h4>Versions Elsewhere</h4>
     </table>
     </div>
 
+    {% cache cache_time package-details-metadata user.is_authenticated %}
```
jelly, line 217: This key is not unique enough and will cache one package and show it one a different package the key needs to be $repo-$arch-$pkgname

PR #527, opened 2024-09-01
anchor to ISO download via mirrors right on top

lcarilla:
indicate that download via HTTP is possible so that people new to arch dont get confused if they dont see the download from mirror option immediately

### DIFF
--- a/templates/public/download.html
+++ b/templates/public/download.html
@@ -25,6 +25,8 @@
     It is intended for new installations only; an existing Arch Linux system
     can always be updated with <code>pacman -Syu</code>.</p>
 
+    <p>Images for installing Arch can be downloaded via <a href="#bittorrent-download">BitTorrent</a> or right here in your browser from one of the <a href="#http-downloads">Arch HTTP(S) mirrors down below</a>.</p>
+
     <ul>
         {% if release.version %}<li><strong>Current Release:</strong> {{ release.version }}</li>{% endif %}
         {% if release.kernel_version %}<li><strong>Included Kernel:</strong> {{ release.kernel_version }}</li>{% endif %}
@@ -52,7 +54,7 @@
     to update your existing system. You may be looking for
     <a href="{% url 'mirrorlist' %}">an updated mirrorlist</a> instead.</p>
 
-    <h3>BitTorrent Download (recommended)</h3>
+    <h3 id="bittorrent-download">BitTorrent Download (recommended)</h3>
 
     <p>If you can spare the bytes, please leave the client open after your
     download is finished, so you can seed it back to others.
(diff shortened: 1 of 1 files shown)


SuperZekes:
This would be a good change, hope the dev team well review this!

templates/public/download.html
```diff
@@ -24,7 +24,7 @@ <h3>Release Info</h3>
     or be <a href="https://wiki.archlinux.org/title/USB_flash_installation_medium">directly written to a USB flash drive</a>.
     It is intended for new installations only; an existing Arch Linux system
     can always be updated with <code>pacman -Syu</code>.</p>
-
+    <p>Images for installing Arch can be downloaded via <a href="#bittorrent-download">BitTorrent</a> or right here in your browser from one of the <a href="#download-mirrors">Arch HTTP(S) mirrors down below</a>.</p>
```
jelly, line 27: No big no from my side but the information is just below, for the Bittorrent download

 The mirror link links to the [mirrors list directly](https://archlinux.org/download/#download-mirrors) which skips all the useful information about verifying an image so that isn't super useful.

What we probably want is to re-order the images but we don't want users to scroll over all the mirrors offering iso images to find out we have docker, cloud and netboot images.

lcarilla: true, we want to make the user aware how to verify the download. adjusted the anchor on top to scroll to the HTTP Direct Downloads section, where the mirrors are linked.

templates/public/download.html
```diff
@@ -91,12 +91,11 @@ <h3>VM images</h3>
...
     <p>Official virtual machine images are available for download on our <a href="https://gitlab.archlinux.org/archlinux/arch-boxes/-/packages">GitLab instance</a>, more information is available in the <a href="https://gitlab.archlinux.org/archlinux/arch-boxes/">README</a>.</p>
 
-    <h3>HTTP Direct Downloads</h3>
+    <h3 id="http-downloads">HTTP Direct Downloads</h3>
 
     <p>In addition to the BitTorrent links above, install images can also be
-    downloaded via HTTP from the mirror sites listed below. Please
+    downloaded via HTTP from the <a href="#download-mirrors">mirror sites listed below</a>. Please
     ensure the download image matches the checksum from the <code>sha256sums.txt</code> or <code>b2sums.txt</code> file linked below.</p>
-
```
jelly, line 99: Is this an accidental removal?

lcarilla: spacing is fixed.

jelly (approved):
I've fixed up the commit message and dropped cssmin.

jelly:
This is better then the previous state, I still think we should revamp the page a bit but that's something for another time.

PR #536, opened 2024-12-07
Improve the "zebra" style of dynamic tables where rows can be dynamically hidden

lahwaacz:
Currently, dynamic tables on archweb may look like this (a todo list with "Only Incomplete" checked; screenshot from a testing instance):

screenshot-2024-12-07@19:21:19

After this commit, it looks like a :zebra: again :smile: 

screenshot-2024-12-07@19:20:25

### DIFF
--- a/sitestatic/archweb.css
+++ b/sitestatic/archweb.css
@@ -754,6 +754,10 @@ table.results {
         text-align: center;
     }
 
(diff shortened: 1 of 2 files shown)


jelly (approved):
Thanks, really wants me to rip out the jQuery table thing we have :(

Issue #520, opened 2024-07-22
ISO 8601 date format


micwoj92:
Hello,

Previously (up until around the [outage](https://old.reddit.com/r/archlinux/comments/1e6n0w1/arch_website_appears_to_be_down/)) packages page used to show dates in ISO 8601 format as can be seen on this snapshot https://web.archive.org/web/20240718044236/https://archlinux.org/packages/
After the outage it uses month day, year format https://web.archive.org/web/20240720124233/https://archlinux.org/packages/

Was there anything changed? Could it be reverted back to ISO 8601, thanks.

jelly:
And https://web.archive.org/web/20231204002243/https://archlinux.org/packages/extra/any/perl-test-most/ and https://archlinux.org/packages/extra/any/perl-test-most/

I don't believe this is a recent change. Maybe more something to do with Django 5.

micwoj92:
Actually it affects entire site. I will update title of the issue.

I don't know if it is new Django version. I had similar issue 2 years ago, but it went away shortly.
https://old.reddit.com/r/archlinux/comments/rknkx5/date_change_on_website/

nl6720:
Quoting https://forum.djangoproject.com/t/datetime-format/30811/10:

> There was a change in the docs between 4.2 and 5.0 that may explain your situation.
> 
> In 4.2, DATETIME_FORMAT:
> 
>> Note that if [USE_L10N](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-USE_L10N) is set to True , then the locale-dictated format has higher precedence and will be applied instead.
> 
> Since the [USE_L10N setting was removed in 5.0 5](https://docs.djangoproject.com/en/5.0/releases/5.0/#features-removed-in-5-0), the docs now read:
> 
>> Note that the locale-dictated format has higher precedence and will be applied instead.
> 
> (In other words, locale-defined formats always take precedence.)
> 
> This means that you either need to change the locale format, or specifically use this format in the template.

nl6720:
I don't know what is the proper way to fix this, but replacing `date:"DATETIME_FORMAT"` with `date:"Y-m-d H:i"` works.

micwoj92:
Sorry for late ping @nl6720, I Must have missed email for this closure.
It is still not 100% ISO 8601 as it was. For example ood flags are not in this format:
https://archlinux.org/packages/extra/x86_64/librime/
Same with /packages and /news:
https://archlinux.org/packages
https://archlinux.org/news/

[...]

PR #537, opened 2024-12-28
Use ISO 8601 date format

nl6720:
Use the `Y-m-d` format everywhere.

Fixes https://github.com/archlinux/archweb/issues/520


jelly (approved):
Thanks!

PR #544, opened 2025-01-18
Run ruff in CI

jelly:


.github/workflows/main.yml
```diff
@@ -9,14 +9,17 @@ jobs:
...
+    - name: Set up Python 3.13
       uses: actions/setup-python@v4
       with:
-        python-version: 3.11
+        python-version: 3.13
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt && pip install -r requirements_test.txt
+        pip install -r requirements.txt && pip install -r requirements_test.txt && pip install ruff
```
lahwaacz, line 19: Why not add it to `requirements_test.txt`?

jelly: Hmm leaving that for when we fully switch to ruff, I hope the relevant E* rules we need will be moved out of preview soonish.

PR #552, opened 2025-02-17
Add `rel=me` to staff members' own website links

wiktor-k:
The `rel=me` is used to indicate [profile equivalence][0] and is used by [Mastodon's Link Verification][1]. If a staff member puts a link to their Fediverse account in the "website" field and cross-links the staff page on their Fediverse account (the link may include `#username` anchor) then the software will display the link as "green" thus giving others an indication that this person is indeed an Arch staff member.

Note that I have *not* ran Archweb locally with this change :sweat_smile: 

I mock-up of how it should look like:

1. Adding the link to Mastodon on archweb:

2. Adding a back-link to the staff site on Mastodon:

3. There's no step 3. It just works and is green and green == good.

CC: @anthraxx with whom I've discussed that at FOSDEM.

[0]: https://microformats.org/wiki/rel-me
[1]: https://docs.joinmastodon.org/user/profile/#verification

### DIFF
--- a/templates/public/developer_list.html
+++ b/templates/public/developer_list.html
@@ -47,7 +47,7 @@
                 </tr><tr>
                     <th>Website:</th>
                     <td>{% if prof.website %}<a itemprop="url" href="{{ prof.website }}"
-                            title="Visit the website for {{ dev.get_full_name }}">
+                            title="Visit the website for {{ dev.get_full_name }}" rel="me">
                             {{ prof.website }}</a>{% endif %}</td>
                 </tr><tr>
                     <th>Occupation:</th>


jelly:
Shouldn't the link be a full link with your mastadon account like `<link rel="me" href="https://fosstodon.org/@archlinux" title="Arch Linux Mastodon">`?

wiktor-k:
> Shouldn't the link be a full link with your mastadon account like `<link rel="me" href="https://fosstodon.org/@archlinux" title="Arch Linux Mastodon">`?

`a href`s work the same way as `link`s with this use-case:

> `<a href="https://social.example.com/@username" rel="me">Follow me on Mastodon!</a>`

([Source](https://docs.joinmastodon.org/user/profile/#verification))

The same thing on GitHub (random example that works: https://github.com/Foxboron) has the following markup: `<a rel="nofollow me" class="Link--primary" style="overflow-wrap: anywhere" href="https://chaos.social/@Foxboron">@Foxboron@chaos.social</a>` and this is enough for Mastodon to "green" the link at: https://chaos.social/@Foxboron

I guess the better approach would be to introduce "social link" but this is far more complex (at least for me :sweat_smile:) as it would require additional DB field.

(I hope I got you right, if not, happy to answer any questions). :wave:

PR #555, opened 2025-02-21
Simplify instructions when using GnuPG for signature verification

wiktor-k:
Since Pierre's signatures contain [Signer's UID packet][UID] GnuPG will [automatically fetch the signing key][T3324] using WKD when using `--auto-key-retrieve`. The filename is assumed to be the same as the signature name minus the `.sig` suffix. Ask the user to confirm that the primary key's fingerprint is correct which makes sure that a correct signing key has been used and not any other that may be in their keyring.

[UID]: https://www.rfc-editor.org/rfc/rfc4880#section-5.2.3.22
[T3324]: https://dev.gnupg.org/T3324

### DIFF
--- a/templates/public/download.html
+++ b/templates/public/download.html
@@ -137,10 +137,13 @@
     With this signing key, verify the signature:
     <pre><code>$ sq verify --signer-file release-key.pgp --signature-file archlinux-{{ release.version }}-x86_64.iso.sig archlinux-{{ release.version }}-x86_64.iso</code></pre>
 
-    Alternatively, using GnuPG, download the signing key from WKD:
-    <pre><code>$ gpg --auto-key-locate clear,wkd -v --locate-external-key {{ release.wkd_email }}</code></pre>
-    Verify the signature:
-    <pre><code>$ gpg --keyserver-options auto-key-retrieve --verify archlinux-{{ release.version }}-x86_64.iso.sig archlinux-{{ release.version }}-x86_64.iso</code></pre>
+    Alternatively, verify the signature using GnuPG:
+    <pre><code>$ gpg --auto-key-retrieve --verify archlinux-{{ release.version }}-x86_64.iso.sig</code></pre>
+
+    And check that the output contains:
+    <pre><code>gpg: Good signature...
+...
+Primary key fingerprint: 3E80 CA1A 8B89 F69C BA57  D98A 76A5 EF90 5444 9A5C</pre><code>
 
     {% cache 600 download-mirrors %}
     <div id="download-mirrors">


dkg:
I'm not a big fan of these changes, for two reasons:

- if someone replaces the signatures with some other signature, then the key retrieval will fetch some other (arbitrary) signing certificate.
- the single-argument form of `gpg --verify` is ambiguous: it can be used either to verify an inline-signed message (that is, an ["OpenPGP Signed Message"](https://www.rfc-editor.org/rfc/rfc9580.html#section-10.3-2.18.1) or a [Cleartext Signed Message](https://www.rfc-editor.org/rfc/rfc9580.html#name-cleartext-signed-message-st)), _or_ it can be used to verify a detached signature whose filename happens to match a known permutation of the signature.  This ambiguity means that if the `wget ….sig` operation fetches a inline-signed message, it's unclear to me what specifically will be verified.

anthraxx:
Honestly dkg has precisely pinpointed why the changes had been made to the current variant in the first place.
So all I can do is second this opinion, please let's not simplify it by losing actual meaningful value in the operations.

wiktor-k:
> if someone replaces the signatures with some other signature, then the key retrieval will fetch some other (arbitrary) signing certificate.


Right. But without the "please verify the fingerprint in the output" the verification may succeed with *any* key that happens to be in the user's keyring so maybe we can have at least this part?

anthraxx:
Can you make it work in a dynamic way like from the release object where the wkd mail comes from? Would be best trying to avoid static text.

dkg:
fwiw, "please verify the fingerprint in the output…" kind of text is exactly the kind of thing that users *will never do* succesfully.

i mean, i might be overstating it a bit just for effect, but the percentage of people who are inclined to follow that step when the rest of it "seems to work fine" is quite low, and the percentage of people who can *actually* successfully follow through on matching 40 digits of hexadecimal is even lower.

What you really want is an automated command that the user can run that either succeeds (quietly) with all the required cryptographic security built in -- or you want a failure with sufficiently explanatory text that they (or their tech support) can figure out what happened.

wiktor-k:
> the percentage of people who can actually successfully follow through on matching 40 digits of hexadecimal is even lower.

I think comparing even several digits raises the bar for the attacker compared to not checking it at all (the "status quo"). But it's not a hill I want to die on so...

> What you really want is an automated command that the user can run that either succeeds (quietly) with all the required cryptographic security built in -- or you want a failure with sufficiently explanatory text that they (or their tech support) can figure out what happened.

Yep, agreed. Is there a `sop` command that we can guide users to? Especially one that satisfies the requirements outlined by @anthraxx ? Or are we comparing instructions that work today but may or may not be followed by users with a mythical :unicorn: tooling? (Genuinely curious and I'm happy to help spec and implement said tooling if it doesn't exist).

dkg:
I think the commands we want are in two parts:

- get the user a copy of the valid cryptographic key material in the form of an OpenPGP certificate
- `sop verify SIGNATURE CERTS < OBJECT`

`sop` doesn't have any certificate fetch functionality (see https://gitlab.com/dkg/openpgp-stateless-cli/-/issues/129 for some discussion), but certainly `wget` would not be an unreasonable way to do it.  From a Linux distro perspective, i can imagine each distro ship the anchor certificates for each other distro in some coordinated way, but i don't think that's done just yet

wiktor-k:
>I think the commands we want are in two parts:
> - get the user a copy of the valid cryptographic key material in the form of an OpenPGP certificate
> - sop verify SIGNATURE CERTS < OBJECT

Okay, this sounds fairly straightforward. I think using `curl` will be slightly better than `wget` since it's in the `core` repository (compared to `wget` being in `extra`).

So that'd be:

```sh
curl -o release-key.pgp 'https://openpgpkey.archlinux.org/.well-known/openpgpkey/archlinux.org/hu/9hy3wi4ewwiicomnjmhewifn6d1gi87i?l=pierre'
rsop verify archlinux-2025.02.01-x86_64.iso.sig release-key.pgp < archlinux-2025.02.01-x86_64.iso
```

(with the appropriate templating in place of raw filenames)

(if that looks good from yours and @anthraxx's side I'm happy to open a PR to replace `sq` with this... if not... well... that's okay too :sweat_smile: )

> From a Linux distro perspective, i can imagine each distro ship the anchor certificates for each other distro in some coordinated way, but i don't think that's done just yet

Yes, although @dvzrv is working on this: https://github.com/uapi-group/specifications/pull/134 I hope this looks good and directly usable from Debian's PoV too. Then we could entirely skip the download step and just use one trust-anchor. Thanks for checking it out! :wave:

dkg:
This seems fine to me, @wiktor-k -- the WKD URL seems a bit ugly to me, perhaps arch wants to set up an HTTP redirection from `https://archlinux.org/release-signing.pgp` to the appropriate WKD URL, so that the `curl` command could be prettier.

Or, if `https://archlinux.org/` is served by an untrustworthy CDN but the `openpgpkey` WKD host is more reliable, you could put the redirect at  `https://openpgpkey.archlinux.org/release-signing.pgp` to still keep the `curl` command a bit prettier.

I'll follow up on the uapi-group spec more there.  thanks for the pointer!

dkg:
for future reference, here's an interesting comparison: https://grapheneos.org/install/cli#obtaining-factory-images GrapheneOS uses SSH keys for signatures and verification, and their verification process is:

- fetch the signing key via `curl`
- fetch the image and detached signature
- verify the detached signature over the image via the signing key

wiktor-k:
> for future reference, here's an interesting comparison: https://grapheneos.org/install/cli#obtaining-factory-images GrapheneOS uses SSH keys for signatures and verification

Agreed that SSH keys are beneficial in a couple of ways:

- the signing key is so small that it fits in a tweet and is easy to copy
- this makes the `allowed_signers` file extremely small and very human readable, just compare [this file](https://github.com/openssh/openssh-portable/blob/master/.git_allowed_signers) to a [file that has the same purpose but is using OpenPGP certs](https://gitlab.com/sequoia-pgp/sequoia/-/blob/main/openpgp-policy.toml?ref_type=heads). This makes writing [cert splitting tools](https://gitlab.archlinux.org/archlinux/archlinux-keyring/-/tree/master/keyring/main/anthraxx/D8AFDDA07A5B6EDFA7D8CCDAD6D055F927843F1C?ref_type=heads) unnecessary since all changes are directly visible in diffs,
- SSH signatures have a "namespace" field, thus making signature reuse harder but ["there is no context separation for signatures in OpenPGP"](https://bugzilla.mozilla.org/show_bug.cgi?id=1953402#c3)
- small technical details on how the final digest is computed makes it easier to perform remote signing in SSH (in SSH and X.509 the data digest is embedded in a SSH framing and *that* is digested and sent to signing, in OpenPGP the packets are stuffed at the end of the data stream), oh, and there's at least a [draft spec for SSH agents](https://datatracker.ietf.org/doc/html/draft-miller-ssh-agent-04)
- since the amount of flexibility is reduced a full-fledged verifier [can be written in 2 days](https://github.com/wiktor-k/ssh-sig)

For the time being Arch is stuck with OpenPGP so I'm looking for a way to make the instructions as straightforward as possible without compromising on security.

Thanks for your input. Deeply appreciated! :bow:

PR #563, opened 2025-04-12
Add the automated install & run method for the WSL image to the Download page

Antiz96:
This automated `wsl --install archlinux` method will only be functional after https://github.com/microsoft/WSL/pull/12818 is merged, so we should wait for that before merging & deploying this PR (I'll remove the draft status once okay).

### DIFF
--- a/templates/public/download.html
+++ b/templates/public/download.html
@@ -95,7 +95,11 @@
 
     <h3>WSL images</h3>
 
-    <p>Official WSL images are available for download on <a href="https://geo.mirror.pkgbuild.com/wsl/latest">mirrors</a>, more information is available in the <a href="https://wiki.archlinux.org/title/Install_Arch_Linux_on_WSL">Wiki</a>.</p>
+    <p>The official WSL image can be installed with the following command (in a PowerShell prompt from a Windows system with WSL 2 installed):</p>
+    <code>wsl --install archlinux</code>
+
+    <p>It is also available for download on <a href="https://geo.mirror.pkgbuild.com/wsl/latest">mirrors</a>.</p>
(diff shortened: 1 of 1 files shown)


templates/public/download.html
```diff
@@ -95,7 +95,10 @@ <h3>VM images</h3>
 
     <h3>WSL images</h3>
 
-    <p>Official WSL images are available for download on <a href="https://geo.mirror.pkgbuild.com/wsl/latest">mirrors</a>, more information is available in the <a href="https://wiki.archlinux.org/title/Install_Arch_Linux_on_WSL">Wiki</a>.</p>
+    <p>The official WSL image is available on <a href="https://geo.mirror.pkgbuild.com/wsl/latest">mirrors</a>. You can run the image with the following command (in a PowerShell prompt from a Winows system with WSL 2 installed):</p>
+    <code>wsl --install archlinux</code>
```
klausenbusk, line 99: Should the sentences be swapped? So the most convenient option is first.
```suggestion
    <p>The official WSL image can be installed with the following command (in a PowerShell prompt from a Windows system with WSL 2 installed):</p>
    <code>wsl --install archlinux</code>
    <p>The official WSL image is also available on <a href="https://geo.mirror.pkgbuild.com/wsl/latest">mirrors</a></p>
```

Antiz96: True. I inspired myself from what was written for [the docker image](https://github.com/archlinux/archweb/blob/master/templates/public/download.html#L89-L90) but your suggestion indeed makes more sense for the WSL image. Thanks!

Antiz96:
This can now be merged & deployed :)

PR #565, opened 2025-04-19
public: Remove Vagrant images

klausenbusk:
We are no longer releasing Vagrant images[1].

[1] https://gitlab.archlinux.org/archlinux/arch-boxes/-/commit/3f7b895725e9fa2f31436234a3b4a4e7f477f69d


jelly:
We should announce this on arch-dev-public or hmm even better write a news announcement.

jelly (approved):

PR #566, opened 2025-05-12
Add `social` field to user's profiles

wiktor-k:
This is a follow up to https://github.com/archlinux/archweb/pull/552

I've went through README.md and nothing failed but there doesn't seem to be any fixtures regarding users.

I'm no Python/Django dev so any help would be very appreciated :sweat_smile:

### DIFF
--- a/devel/models.py
+++ b/devel/models.py
@@ -40,6 +40,9 @@ class UserProfile(models.Model):
     website = models.URLField(max_length=200, null=True, blank=True)
     website_rss = models.URLField(max_length=200, null=True, blank=True,
                                   help_text='RSS Feed of your website for planet.archlinux.org')
+    social = models.URLField(max_length=200, null=True, blank=True,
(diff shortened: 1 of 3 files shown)


devel/models.py
```diff
@@ -40,6 +40,7 @@ class UserProfile(models.Model):
     website = models.URLField(max_length=200, null=True, blank=True)
     website_rss = models.URLField(max_length=200, null=True, blank=True,
                                   help_text='RSS Feed of your website for planet.archlinux.org')
+    social = models.URLField(max_length=200, null=True, blank=True)
```
lahwaacz, line 43: Maybe add some `help_text` like above, otherwise it may not be clear that it expects a URL when the user sees "Social:" in the web form

wiktor-k: Right! :man_facepalming:

jelly (approved):
Looks good!

Issue #550, opened 2025-02-10
"Only Mine" filter for package signoffs list


polyzen:
Similar to the one for todo lists.

PR #571, opened 2025-06-09
signoffs: add "Only Mine" filter

polyzen:
Implements https://github.com/archlinux/archweb/issues/550


jelly (approved):
Looks good, but do note that this only applies to packages you packaged and not packages you maintain :)

PR #577, opened 2025-07-24
Fix typo in venv instructions

svenstaro:
This needs to be `venv` and not `env` since the instructions above call it `venv`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ packages, you will probably want the following:
 
 2. Activate the virtualenv.
 
-        source ./env/bin/activate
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ packages, you will probably want the following:
 
 2. Activate the virtualenv.
 
-        source ./env/bin/activate
+        source ./venv/bin/activate
```
jelly, line 41: `python -m venv env` creates a virtualenv called `env`, `venv` is the python module name.

PR #578, opened 2025-08-29
Fix Django deprecation warnings

benmezger:
PR fixes Django deprecation warnings.


devel/forms.py
```diff
@@ -31,6 +32,11 @@ def clean(self):
             raise forms.ValidationError('Passwords do not match.')
         return self.cleaned_data
 
+def _urlfields_assume_https(db_field, **kwargs):
+    if isinstance(db_field, URLField):
+        kwargs["assume_scheme"] = "http"
+    return db_field.formfield(**kwargs)
```
benmezger, line 38: We could avoid doing this and default to `https` instead using `FORMS_URLFIELD_ASSUME_HTTPS` in `settings.py`. So if we get `archlinux.org` it becomes `https://archlinux.org`. Currently, it defaults to `http`. At the moment, we [default](https://github.com/django/django/blob/23a853821b75787d77016811881220ec6f57310a/django/forms/fields.py#L786) to `http`, so I kept the behaviour.

jelly: This is for the URLField, so making this default to https is imo fine. So yes please change it to `FORMS_URLFIELD_ASSUME_HTTPS=True` in `settings.py` with an explanation that this is temporary until django 6.0 makes it the default.

jelly: I don't feel it is worth the added code and complexity to silence a deprecation warning, just setting `FORMS_URLFIELD_ASSUME_HTTPS = True` in `settings.py` would be good enough for me. Even though it emits a deprecation warning.

jelly (changes requested):

devel/forms.py
benmezger: Agreed. Pushed force the changes + updated description.

jelly:
Thanks!

PR #579, opened 2025-08-30
Replace `pip` with `uv`

benmezger:
I've been using `UV` in `archweb` to handle dependencies and `venv` management a little easier. Given that, I am opening a PR with the changes I made, in case you want to migrate from `pip` to `UV`.

I will take care of the infrastructure PR, so if one wants to merge this, it's better to wait for the infra PR.


.gitlab-ci.yml
benmezger, line 1: @jelly is this still being used somewhere?

.python-version
```diff
@@ -0,0 +1 @@
+3.13.0
```
jelly, line 1: Can't this be part of pyproject.toml?

requirements_prod.txt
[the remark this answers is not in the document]
jelly: Is that UV specific or an accepted Python PEP?

pyproject.toml
```diff
@@ -31,9 +31,51 @@ dev = [
     "pytest>=8.4.1",
     "pytest-cov>=6.2.1",
     "pytest-django>=4.11.1",
+    "ruff>=0.12.11",
```
jelly, line 34: venv managed ruff :(

main/storage.py
```diff
@@ -1,10 +1,9 @@
+import cssmin
 import jsmin
 from django.contrib.staticfiles.storage import ManifestStaticFilesStorage
 from django.core.files.base import ContentFile
 from django.utils.encoding import smart_str
 
-import cssmin
```
jelly, line 6: `ruff check .` is happy for me, why is the uv used ruff not?

Makefile
```diff
@@ -18,15 +18,15 @@ test-pdb: test
 .PHONY: test lint coverage
 
 lint:
-	flake8 --extend-exclude "*/migrations/,local_settings.py" devel main mirrors news packages releng templates todolists visualize *.py
+	uv run flake8 --extend-exclude "*/migrations/,local_settings.py" devel main mirrors news packages releng templates todolists visualize *.py
```
jelly, line 21: should re-evaluate if ruff already has the formatting rules, or ugh `ruff format` I guess.

.gitlab-ci.yml
jelly: No, but eventually we should move to GitLab I guess :)

jelly (commented):
Cool, I'll have to give this a test spin. I am not entirely sure if I'll like uv (never used it before).

This also calls for us switching to dependabot, but not sure if that can bump the uv version in the github actions workflow.

.python-version
benmezger: It can. `uv` creates this by default, given some Python toolings read this file. I will remove this and keep it in `pyproject.toml`, then update `workflow` to revert to its original version selection.

main/storage.py
benmezger: Could it be that you are running a different version of `ruff`? We are installing the latest in the CI:

https://github.com/archlinux/archweb/blob/59229f280bb92fae770bc1efe690d82b71cd6e34/.github/workflows/main.yml#L19

This is one of the benefits of managing dependencies with `uv`, since we are now locking a particular version and making sure we always run that version, without having a separate `requirements` for each group.

requirements_prod.txt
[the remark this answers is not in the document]
benmezger: Looks like it was left [out of scope](https://peps.python.org/pep-0621/#recommend-that-tools-put-development-related-dependencies-into-a-dev-extra) when PEP-0621 was proposed. Poetry and UV move dependencies into groups, a default one that is always installed, along with optional [groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups) of dependencies that we can install.

`uv sync` installs dependencies + dev dependencies
`uv sync --no-dev` install project dependencies without any development
`uv sync --group prod` installs dependencies + prod dependencies
etc. 

This makes it easy for us to manage different scope of dependencies within the project

benmezger:
> This also calls for us switching to dependabot, but not sure if that can bump the uv version in the github actions workflow.

@jelly it [can](https://github.blog/changelog/2025-03-13-dependabot-version-updates-now-support-uv-in-general-availability/): 

```yaml
version: 2
enable-beta-ecosystems: true
updates:
  - package-ecosystem: "uv"
    directory: "/"
    schedule:
      interval: "daily"
```

.python-version
benmezger: Addressed in [`6634c9b` (#579)](https://github.com/archlinux/archweb/pull/579/commits/6634c9b341b841ead2f39c372e3fe7cf32121fb7)

benmezger:
@lahwaacz I think Arch's Gitlab uses `renovate`, but I don't think it [supports](https://docs.renovatebot.com/modules/manager/) `uv` yet.

jelly (approved):
Sorry for the long delay, finally picked it up fixed some small mistakes and will merge it now.

Thanks!

Issue #582, opened 2025-09-16
Package links in ToDo lists are broken


lfos:
Package links seem to have an extra `s%22` appened to the URL; e.g., on https://archlinux.org/todo/cuda-13/, the adios2 entry links to https://archlinux.org/packages/extra/x86_64/adios2/s%22.

christian-heusel:
Yeah this seems to be a leftover from #578 / https://github.com/archlinux/archweb/commit/0cf24055d5181fa516c8182a29f1deef7652b725:
https://github.com/archlinux/archweb/blob/fb9b330535412fd216a6ad09efdbcc5db2903b59/todolists/templatetags/todolists.py#L16

lahwaacz:
I've opened https://github.com/archlinux/archweb/pull/583 to fix this.

benmezger:
Yup, I probably missed this when self-reviewing. Like I said in the PR, I will push a follow-up PR in the EOW with some tests, to make sure we catch such issues.

PR #583, opened 2025-09-16
Fix format string for package links in todo lists

lahwaacz:
Fixes 0cf24055d5181fa516c8182a29f1deef7652b725

Fixes https://github.com/archlinux/archweb/issues/582


benmezger (approved):
Good catch, thanks for taking care of it. 
I will work on a follow-up PR to add some tests to these, so we prevent this from happening next time.

christian-heusel (approved):

jelly (approved):
Thanks!

PR #591, opened 2025-10-05
Fix URL formatting for PGP key lookup

felixonmars:
The URL currently returns a "Bad Request" error without the change here.


jelly:
Thanks!

PR #595, opened 2025-10-28
Developer -> Packager

eworm-de:
Actually one a subset is developer, but all are packager.


jelly (approved):
Sure! (Or Maintainer)

eworm-de:
Fine as well...

PR #596, opened 2025-10-31
Switch the WSL image download link to the new Fastly CDN mirror

Antiz96:
We are now using / promoting the new Fastly CDN mirror as the default one for releng projects (see https://gitlab.archlinux.org/archlinux/archlinux-wsl/-/issues/17)


Antiz96:
cc @mhegreberg

mhegreberg:
Lgtm

PR #600, opened 2025-12-18
Correct number of AUR build scripts in about.html

ismailarilik:
Updated the number of build scripts in the AUR from 49,000 to 100,000.

This is not a big issue but necessary to be competitive IMHO.


ismailarilik:
Anyone here? This is a long-waiting PR which should be accepted or closed IMHO. @jelly What is your opinion?

jelly:
> Anyone here? This is a long-waiting PR which should be accepted or closed IMHO. @jelly What is your opinion?

I'll close it as we can just keep bumping the number and its not necessary to be competitive.

PR #661, opened 2026-04-11
Fix duplicate maintainers on package pages (#427)

### LINKED ISSUE #427: Maintainer doubled on specific package
https://archlinux.org/packages/community/any/python-qt.py/

image

m1rm:
Fixes: #427 

This pull requests appends .distinct() to the User queryset in Package.maintainers so each user is returned at most once, regardless of duplicate relation rows.

### DIFF
--- a/main/models.py
+++ b/main/models.py
@@ -177,7 +177,8 @@ class Package(models.Model):
         if self._maintainers is None:
             self._maintainers = User.objects.filter(
                 package_relations__pkgbase=self.pkgbase,
-                package_relations__type=PackageRelation.MAINTAINER)
+                package_relations__type=PackageRelation.MAINTAINER,
+            ).distinct()
         return self._maintainers
 
     @maintainers.setter


christian-heusel:
Whats an example package where this bug occurs? 😊

m1rm:
> Whats an example package where this bug occurs? 😊

The bug still appears here: https://archlinux.org/packages/extra/any/python-qt.py.

I also tried reproducing it locally using the commands from the current README.md step no. 7 to load real prod data locally, but _python-qt_ then was not found when running locally. So I assume it might be worth a thought providing some updated dumps (which would be a separate issue/todo ofc).

I also think the distinct is rather a "workaround" then a real fix. I assume we would have to check prod tables and maybe do an according migration to clean up data actually causing the bug in the first place (out of my reach however).

m1rm:
> after I stopped overseeing that one sentence in the README, that I can import other data easily, I tried:
wget http://mirrors.kernel.org/archlinux/extra/os/any/extra.files.tar.gz && uv run ./manage.py reporead --filesonly any extra.files.tar.gz -> 404 (typo in url, any is not an architecture)
wget http://mirrors.kernel.org/archlinux/extra/os/x86_64/extra.files.tar.gz && uv run ./manage.py reporead --filesonly x86_64 extra.files.tar.gz    works but when I - for example - search for 7zip locally, it is not found.
sorry for closing and opening - I am switching platforms to much and had a "ctrl+ Enter" for newline, which is not needed in Github.

**Update**: I used the geo mirror to correctly import core & extra. Yet, python-qt is still missing locally. I am also not sure if I can reproduce it locally bc. the packagedata are of no help regarding corrupted data in archweb's DB.

Is there any way I can obtain a db dump from archweb production? (might not be that easy/possible regarding sensitive data/data privacy). Without data insight, it seems near impossible to fix the root cause

PR #660, opened 2026-04-11
docs: add git submodule setup to documentation

m1rm:
Change:
This pull requests adds a step in the README.md on how to setup git submodules (archlinux common style).

Issue:
If not executed when setting up the project for the first time, former step no.4,  `uv run ./manage.py migrate`, fails with errors.

Motivation:
Make it easier for new contributors to contribute/lower the hurdle.

PR #662, opened 2026-04-11
Add pkgbases export endpoint to developer reports

benmezger:
This PR adds pkgbases export endpoint to developer reports, mirroring the existing todolist endpoint.

screen

This solves #659.

### DIFF
--- a/devel/views.py
+++ b/devel/views.py
@@ -266,6 +266,22 @@ def change_profile(request):
 
 
 @login_required
+def report_pkgbases(request, report_name: str, username: str | None = None) -> HttpResponse:
+    report = {report.slug: report for report in available_reports()}.get(report_name, None)
+    if report is None:
+        raise Http404
+
+    packages = Package.objects.normal()
+    if report.slug in ('uncompressed-man', 'uncompressed-info'):
+        packages = report.packages(packages, username)
+    else:
+        packages = report.packages(packages)
+
+    pkgbases = sorted({pkg.pkgbase for pkg in packages})
+    return HttpResponse('\n'.join(pkgbases), content_type='text/plain')
+
+
+@login_required
 def report(request, report_name, username=None):
     available = {report.slug: report for report in available_reports()}
     report = available.get(report_name, None)
--- a/devel/views.py
+++ b/devel/views.py
@@ -265,6 +265,27 @@ def change_profile(request):
                    'profile_form': profile_form})
 
 
+def get_report_packages(report, username):
+    packages = Package.objects.normal()
+    if report.slug in ('uncompressed-man', 'uncompressed-info'):
+        packages = report.packages(packages, username)
+    else:
+        packages = report.packages(packages)
+
+    return packages
+
+
+@login_required
+def report_pkgbases(request, report_name: str, username: str | None = None) -> HttpResponse:
+    report = {report.slug: report for report in available_reports()}.get(report_name, None)
+    if report is None:
+        raise Http404
+
+    packages = get_report_packages(report, username)
+    pkgbases = sorted({pkg.pkgbase for pkg in packages})
+    return HttpResponse('\n'.join(pkgbases), content_type='text/plain')
+
+
 @login_required
 def report(request, report_name, username=None):
     available = {report.slug: report for report in available_reports()}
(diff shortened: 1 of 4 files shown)


devel/views.py
```diff
@@ -290,6 +308,8 @@ def report(request, report_name, username=None):
 
     arches = {pkg.arch for pkg in packages}
     repos = {pkg.repo for pkg in packages}
+    svn_roots = sorted({repo.svn_root for repo in repos})
```
benmezger: `svn_roots` is  derived from the `repos` actually present in the reports packages

christian-heusel: Ooof, that should really not be called `repo.svn_root` anymore (we migrated to git a long time ago), but I guess that refactoring is to be done in a separate change and unrelated to this one 😬 

https://github.com/archlinux/archweb/blob/459a71db353f7569a322a5b075885dfab6bcc09a/main/models.py#L77-L78

README.md
```diff
@@ -33,6 +33,7 @@ probably want the following:
 
 # Testing Installation
 
+1. Pull git submodules: `git submodule update --init --recursive`
```
christian-heusel: I think a more extensive version of this change is already staged in #660

devel/views.py
benmezger: Yeah, that's something I've noticed while writing this code. I've kept `svn_root` for now just to keep the current standard and not diverge from it, but I will take care of renaming this in a following PR and making sure we rename it to `git_root` or something similar.

README.md
benmezger: Ah right, I think I missed seeing that PR. I've dropped the README commit :)

devel/views.py
christian-heusel: Potentially even something like `pkgbase` would work, as this is how we map it generally (with a few exceptions): like `ls++` or `tree` 
https://github.com/archlinux/archweb/blob/459a71db353f7569a322a5b075885dfab6bcc09a/main/utils.py#L156-L165

But the exact naming is something which most likely the archweb maintainers should give input on 😊

benmezger: I've opened a [PR](https://github.com/archlinux/archweb/pull/663) initially setting to `git_root` and linked up you comment as suggestion. Lets see what they suggest :)

jelly: @christian-heusel not sure why you pull `gitlab_project_name_to_path` into the discussion, the `svn_root` is either `community` or `packages` historically. This simply comes from the Developer / Trusted Users split in svn repositories. This is no longer relevant since we moved to GitLab which has one package repository instead of one svn repository.

So I believe we can just refactor this all away:

Image

Show community/packages here is imo not needed.

jelly (changes requested):
Since `svn_root` is meaningless in our new Git packaging repository workflow lets not depend on it for new code and instead make the endpoint `/pkgbases` without the `svn_root`. We can later refactor the To-Do list view to link to `pkgbases` instead of `community` or `packages`.

benmezger:
> Since `svn_root` is meaningless in our new Git packaging repository workflow lets not depend on it for new code and instead make the endpoint `/pkgbases` without the `svn_root`. We can later refactor the To-Do list view to link to `pkgbases` instead of `community` or `packages`.

@jelly that makes sense. With regards to the HTML changes, that was done to keep the same visual functionality as the todo page, but since we will drop svn from the endpoint, that makes sense.

I've updated the code accordingly.

devel/views.py
```diff
@@ -265,6 +265,22 @@ def change_profile(request):
...
+def report_pkgbases(request, report_name: str, username: str | None = None) -> HttpResponse:
+    report = {report.slug: report for report in available_reports()}.get(report_name, None)
+    if report is None:
+        raise Http404
+
+    packages = Package.objects.normal()
+    if report.slug in ('uncompressed-man', 'uncompressed-info'):
+        packages = report.packages(packages, username)
+    else:
+        packages = report.packages(packages)
```
jelly, line 278: If you are up to it, would be nice to put this in a separate function and commit and refactor the existing `report` function to use that so we aren't copying logic around.

jelly (changes requested):
This still needs some UI to make the link discover-able I would copy the design of the existing ToDo list but just make it `Link to pkgbases`

jelly (approved):

benmezger:
@jelly thanks for taking care. Was busy with life and planned this for later 🙇

PR #663, opened 2026-04-11
Drop `svn_root` and replace `todolists` with `pkgbases`

benmezger:
PR drops `svn_root` column, as its now meaningless with our Git repository workflow. 

This could potentially be a breaking change in case another application that reads the same database relies on the `svn_root` column or some external API depends on that name.


main/models.py
```diff
@@ -74,8 +74,8 @@ class Repo(models.Model):
         default=1, help_text="Flyspray project ID for this repository.")
     bugs_category = models.SmallIntegerField(
         default=2, help_text="Flyspray category ID for this repository.")
-    svn_root = models.CharField(
-        max_length=64, help_text="SVN root (e.g. path) for this repository.")
+    git_root = models.CharField(
```
benmezger: I wonder if we should choose a more generic name rather than `git_root`. `repo_root` perhaps?

benmezger: Christian proposal on this https://github.com/archlinux/archweb/pull/662#discussion_r3067950923

> Potentially even something like pkgbase would work, as this is how we map it generally (with a few exceptions): like ls++ or tree
> https://github.com/archlinux/archweb/blob/459a71db353f7569a322a5b075885dfab6bcc09a/main/utils.py#L156-L165
> But the exact naming is something which most likely the archweb maintainers should give input on 😊

jelly (changes requested):
I think its better to drop `svn_root` altogether and replace it in the todo's with a link to `pkgbases`. What might be better to split this into core/extra but honestly lets make that a follow up.

main/fixtures/repos.json
```diff
@@ -1,146 +1,134 @@
...
+            "bugs_category": 10,
+            "staging": true,
             "name": "Core-Staging",
-            "bugs_project": 1, 
-            "svn_root": "packages", 
+            "bugs_project": 1,
             "testing": false
         }
     }
-]
```
jelly, line 146: All kinds of unintended trailing space removal and missing newline at the end? Please just drop `svn_root` only.

templates/todolists/view.html
```diff
@@ -32,10 +32,7 @@ <h2>{{ list.kind_str | title }} Todo List {{ list.name }}</h2>
     </div>
 
     <div class="todo-pkgbases">
-        <p>Link to lists of pkgbase values:</p>
-        <ul>{% for svn_root in svn_roots %}
-            <li><a href="pkgbases/{{ svn_root }}/">{{ svn_root }}</a></li>
-        {% endfor %}</ul>
+        <p><a href="pkgbases/">List of pkgbase values</a></p>
```
jelly, line 35: 👍  nice, same can be used in the devel reports.

jelly (changes requested):
Just needs some small minor fixes and its good to merge.

jelly (approved):

benmezger:
@jelly thanks for taking care. Was busy with life and planned this for later 🙇

PR #671, opened 2026-04-30
Fix: frontpage news codeblock linebreaks & add news fixtures

### LINKED ISSUE #573: Front page news lose line breaks in code blocks
My recent news post on [its own page](https://archlinux.org/news/linux-firmware-2025061312fe085f-5-upgrade-requires-manual-intervention/) looks fine, but on [the front page](https://archlinux.org/) the code blocks had their line breaks collapsed into spaces.

The markup is fine (it's a `<code>` block in a `<pre>` block). The actual line breaks are missing.

m1rm:
> [!TIP]
> This pull request aims at fixing missing line breaks in frontpage news code blocks. Additionally, for local development, news fixtures for the frontpage news section were added.

#### Issue linebreak loss
<details><summary> vendor code details</summary>
The `truncatewords_html` filter uses the TruncateWordsHTMLParser
-> The filter is in: `.venv/lib/python3.14/site-packages/django/template/defaultfilters.py`
-> The Truncator is in: `.venv/lib/python3.14/site-packages/django/utils/text.py`

```
class TruncateWordsHTMLParser(TruncateHTMLParser):
    def process(self, data):
        data = re.split(r"(?<=\S)\s+(?=\S)", data)
        output = escape(" ".join(data[: self.remaining]))
        return data, output
```
-> **I think there the \n is lost because of the \S and \s usage (see https://docs.python.org/3/library/re.html)**
</details>

#### Potential fixes
**Using the linebreak filter**
Commit: f5435909b055a00a252a5d97b47d265d3756bea8 -> might not be a good choice, see comment screenshot

**Using a custom filter**
The custom filter is based on the builtin filter and only changes the behavior for `<pre>` blocks
Commit: d05be818d9d4efe53fc0a9a52664d6b6627a5d59

#### News fixtures
In order to debug this issue locally, fixtures for news are added (they are fully made up, I am not emotionally attached to the content so feel free to change it).

#### Tests
I added a test for newline preservation. Not sure if we need it but we can always remove it.

Fixes https://github.com/archlinux/archweb/issues/573

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ probably want the following:
         uv run ./manage.py loaddata devel/fixtures/*.json
         uv run ./manage.py loaddata mirrors/fixtures/*.json
         uv run ./manage.py loaddata releng/fixtures/*.json
+        uv run ./manage.py loaddata news/fixtures/*.json
 7. Use the following commands to start a service instance
 
         uv run ./manage.py runserver
--- a/main/templatetags/htmltruncate.py
+++ b/main/templatetags/htmltruncate.py
@@ -1,0 +1,55 @@
+import re
+
+from django import template
+from django.template.defaultfilters import stringfilter
+from django.utils.html import escape
+from django.utils.text import TruncateWordsHTMLParser
+
+register = template.Library()
+
+
+class PrePreservingTruncateWordsHTMLParser(TruncateWordsHTMLParser):
+    """
+    Like Django's TruncateWordsHTMLParser, but text inside <pre> keeps its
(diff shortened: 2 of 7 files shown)


m1rm:
converted to draft again bc. I am checking if the "linebreak" filter might add too many linebreaks. @jelly @christian-heusel because it is applied to all the content (not only codeblocks)

m1rm:
### Appearance when using the linebreak filter:

Screenshot From 2026-04-30 19-41-51


### Appearance when using the custom filter:

Screenshot From 2026-04-30 19-42-18

Issue #199, opened 2019-02-21
Create a real REST API for accessing package data


jelly:
Currently we have no real official REST API or consistent JSON API. We need to figure out what data we want to have easily query'able for users.

Current "API":
* http://www.archlinux.org/packages/search/json/?q=$search term
* https://www.archlinux.org/packages/$repo/$arch/$pkgname/json/
* https://www.archlinux.org/packages/$repo/$arch/$pkgname/files/json/
* https://www.archlinux.org/todo/$todoname/json
* https://www.archlinux.org/packages/signoffs/json/
* https://www.archlinux.org/mirrors/status/json/
* https://www.archlinux.org/mirrors/status/tier/1/json/ 
* https://www.archlinux.org/mirrors/locations/json/
* https://www.archlinux.org/master-keys/json/
* https://www.archlinux.org/releng/releases/json/

What's missing
* A generic way to create a token and for example mark signoffs securely. arch-signoff now relies on doing http basic auth
* Todo list overview endpoint https://www.archlinux.org/todo/json
* Documentation
* Performance

Bezmehrabi:
may i help you with Django restframework? and maybe a little more detail about what you want?

felipec:
From [task 57137](https://bugs.archlinux.org/task/57137), it would be nice to be able to query the last version of a bunch of packages, AUR provides such interface:

https://aur.archlinux.org/rpc/?v=5&type=info&arg[]=spotify&arg[]=pacaur

Right now the query has to be done per-package, which is extremely inefficient.

Other APIs provide similar queries, where you can get a specific attribute for a bunch of IDs, some limit the number, like 20.

For example:

http://www.archlinux.org/packages?ids=spotify,pacaur&attributes=version

Bezmehrabi:
it's been a long time..., but I'm still willing to do it because I love the arch community :)
I can do it with either rust or python.
let me know which one you guys(maintainers) prefer?

felipec:
I am willing to help as well. I've been writing REST APIs lately.

PR #674, opened 2026-05-09
feat(api): add Django Ninja with releng releases endpoint

inglor:
Introduces the api/ Django app with django-ninja==1.6.2. Mounts NinjaAPI at /api/ with OpenAPI schema at /api/openapi.json and Swagger UI at /api/docs/. Implements the first route:

  GET /api/v1/releng/releases/ -> ReleasesSchema

Pydantic schema mirrors ReleaseJSONEncoder output. 10 tests covering endpoint behaviour and OpenAPI schema contract.

Add decorator for deprecation HTTP header RFC 8594

Closes part of #199.

### DIFF
--- a/api/schemas/releng.py
+++ b/api/schemas/releng.py
@@ -1,24 +1,23 @@
 from datetime import date, datetime
-from typing import Optional
 
 from ninja import Schema
 
 
 class ReleaseSchema(Schema):
     version: str
-    kernel_version: Optional[str] = None
+    kernel_version: str | None = None
     release_date: date
     available: bool
     info: str
-    iso_url: Optional[str] = None
-    magnet_uri: Optional[str] = None
-    torrent_url: Optional[str] = None
-    md5_sum: Optional[str] = None
-    sha1_sum: Optional[str] = None
-    sha256_sum: Optional[str] = None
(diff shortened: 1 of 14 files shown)


api/schemas/releng.py
```diff
@@ -0,0 +1,29 @@
...
+from ninja import Schema
+
+
+class ReleaseSchema(Schema):
+    version: str
+    kernel_version: Optional[str] = None
+    release_date: date
+    available: bool
+    info: str
+    iso_url: Optional[str] = None
```
jelly: The project uses modern enough python to write `str | None`

jelly (commented):
Note, that you'll need a rebase locally if you pulled stuff from master and run django migrations due to svn_root removal.

This is super interesting, especially since we want to be allow users to sign off packages using an API or marking packages from a todolist.

For that we need authentication which ninja seems to [support](https://django-ninja.dev/guides/authentication/). So I would be interested in experimenting with that.

api/schemas/releng.py
inglor: Done!

jelly (approved):
Lets merge this, we can extend it later.

PR #675, opened 2026-05-09
feat(#590): add "Testing Version" column to todo lists

### LINKED ISSUE #590: Add a "Testing Version" column to todo lists
Before moving a rebuild from the staging to testing repos, it'd be good to be able to quickly spot potential overlaps with other packages that haven't been moved to [core]/[extra] yet.

m1rm:
This PR adds the column  "Testing Version" to /todos/:a_list containing the version of the package in the Testing Repo, if present.

Additionally, fixtures and tests are added.

Closes https://github.com/archlinux/archweb/issues/590

### DIFF
--- a/todolists/utils.py
+++ b/todolists/utils.py
@@ -39,21 +39,34 @@ def get_annotated_todolists(incomplete_only=False):
     return lists
 
 
+def _attach_repo_version(packages, list_id, kind):
+    '''Look for related packages in repos flagged by kind and attach them.
+
+    kind is a Repo boolean field name used as repo__{kind}=True. Only
+    'staging' and 'testing' are used; they set package.staging or
+    package.testing for the todolist templates.
+    '''
+    pkgnames = TodolistPackage.objects.filter(
+        todolist_id=list_id).values('pkgname')
+    related_pkgs = Package.objects.normal().filter(
+        **{f'repo__{kind}': True},
+        pkgname__in=pkgnames,
+    )
+    lookup = {(p.pkgname, p.arch): p for p in related_pkgs}
+
+    for package in packages:
+        setattr(package, kind, lookup.get((package.pkgname, package.arch)))
+
+
 def attach_staging(packages, list_id):
     '''Look for any staging version of the packages provided and attach them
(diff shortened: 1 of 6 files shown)


todolists/utils.py
```diff
@@ -56,4 +56,22 @@ def attach_staging(packages, list_id):
...
+    '''Look for any testing version of the packages provided and attach them
+    to the 'testing' attribute on each package if found.'''
+    pkgnames = TodolistPackage.objects.filter(
+        todolist_id=list_id).values('pkgname')
+    testing_pkgs = Package.objects.normal().filter(repo__testing=True,
+                                                   pkgname__in=pkgnames)
+    # now build a lookup dict to attach to the correct package
+    lookup = {(p.pkgname, p.arch): p for p in testing_pkgs}
+
+    annotated = []
```
christian-heusel: This seems to be unused, why is it there? 🤔 

I see that it's also part of the logic in `attach_staging()`, but maybe you have some context of it now that you worked on `attach_testing()`.

todolists/utils.py
```diff
@@ -56,4 +56,22 @@ def attach_staging(packages, list_id):
...
+    for package in packages:
+        in_testing = lookup.get((package.pkgname, package.arch), None)
+        package.testing = in_testing
+
+    return annotated
```
christian-heusel: Would it be an option to refactor `attach_staging()` to use a parameter to keep the overall code more DRY?

christian-heusel (commented):
Thanks a lot for working on this! 🤗 🚀  I have two small questions and will defer to the project maintainers for the overall verdict! 😄

todolists/utils.py
m1rm: I think yes, I'll update the PR shortly

todolists/utils.py
m1rm: ah true. I only "copied" the behavior of "attach_staging". I think it is a leftover, I will remove it 👍

m1rm: the "real magic" happens with the lookup dict regarding performance and in the loop afterwards the data is "attached" (the versions of the packages)

PR #676, opened 2026-05-09
ci/cd: prevent multiple workflows for the same code

m1rm:
This change prevents multiple runs for the same code, which can happen if someone pushes changes quick enough after a previous push.

See also: https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,6 +2,10 @@ name: Github-Actions
 
 on: [push, pull_request]
 
+concurrency:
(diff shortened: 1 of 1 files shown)


jelly (commented):
This feels like a very niche situation, this repository has very low activity so it doesn't feel worth it to optimise for it.

m1rm:
> This feels like a very niche situation, this repository has very low activity so it doesn't feel worth it to optimise for it.

I understand :) just thought it's a nice small thing but yeah, it is propably not necessary

PR #681, opened 2026-05-15
Add link to the Developer Manual

lahwaacz:
There are not many pages yet but we need to improve the visibility somehow...


jelly (approved):
Thanks!

PR #682, opened 2026-05-15
Compress flagged packages on the dashboard

lahwaacz:
Out-of-date flagging and version bumping is related to pkgbase rather than pkgname, it does not make sense to duplicate the same information on multiple rows in the dashboard.

Before:

screenshot-2026-05-15@22-24-40

After:

screenshot-2026-05-15@22-24-03


jelly (approved):
Thanks!

PR #691, opened 2026-05-24
fix: display better pkg link for flagged group of split packages

lahwaacz:
Fixes the issue pointed out in this comment:
https://github.com/archlinux/archweb/commit/2e28e98927cf42c15e12e304d5f8ca783850bf0d#commitcomment-185933020

When a split package has pkgbase name that is alphabetically after a pkgname of another package, we need a sorting trick to display a link to the correct package. This follows the behavior of the signoff table, only without the PackageSignoffGroup class.


lahwaacz:
cc @eworm-de

eworm-de:
Great, thanks a lot!

jelly (approved):
Sure!

PR #693, opened 2026-05-25
Add rsync mirror note to download page

YukiCodepth:
## Summary

Add a short note mentioning rsync support on the download page.

## Changes

- Add a sentence informing users that some Arch Linux mirrors support rsync
- Place the note below the HTTP direct downloads section

## Verification

- Verified the new note renders correctly on the downloads page
- Confirmed formatting matches surrounding content

### DIFF
--- a/templates/public/download.html
+++ b/templates/public/download.html
@@ -100,6 +100,8 @@
     downloaded via HTTP from the <a href="#download-mirrors">mirror sites listed below</a>. Please
     ensure the download image matches the checksum from the <code>sha256sums.txt</code> or <code>b2sums.txt</code> file linked below.</p>
 
+    <p>Many Arch Linux mirrors also support rsync.</p>
+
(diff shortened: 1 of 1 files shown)


templates/public/download.html
```diff
@@ -100,6 +100,8 @@ <h3 id="http-downloads">HTTP Direct Downloads</h3>
     downloaded via HTTP from the <a href="#download-mirrors">mirror sites listed below</a>. Please
     ensure the download image matches the checksum from the <code>sha256sums.txt</code> or <code>b2sums.txt</code> file linked below.</p>
 
+    <p>Many Arch Linux mirrors also support rsync.</p>
```
jelly, line 103: The section above is about downloading images/artifacts that has nothing to do with rsync.

jelly (changes requested):

Issue #586, opened 2025-09-18
Flagging an old version should not flag newer versions


heftig:
When, e.g., `gnome-backgrounds 48.2.1-1` in `[extra]` gets flagged out-of-date, this should not also flag `gnome-backgrounds 49.0-1` in `[gnome-unstable]`.

I guess flags should only spread to "split" packages of the same pkgver.

m1rm:
I did some refactoring regarding this in the linked pull request. I am not 100% firm regarding the different repositories so I focused on version constraints - I think it is an improvement but only "half" of is wanted when I read your comment @jelly . 

I'd suggest to implement the behavior of "flag testing -> flags core; flag core -> does not flag testing" in a subsequent pull request (which I will be happy to work on, too).

PR #711, opened 2026-06-17
fix(#586): flagging a pkgver out of date no longer flags newer versions

m1rm:
This pull request:

- adds a version filter when flagging a package out of date: `alpm.vercmp(peer.full_version, self.full_version) <= 0`. This is the core fix for #586 
- refactors the core logic of flagging into `flag_peers()` which makes the code a bit more DRY
- adds a test for flagging packages behavior

Relates to: https://github.com/archlinux/archweb/issues/586

### DIFF
--- a/main/models.py
+++ b/main/models.py
@@ -373,6 +373,35 @@ class Package(models.Model):
             repo__staging=self.repo.staging,
             pkgbase=self.pkgbase).exclude(id=self.id)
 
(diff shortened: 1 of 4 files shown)


m1rm:
I think this PR and the associated bug can be closed because manually flagging is disabled? https://github.com/archlinux/archweb/pull/719 cc @jelly

jelly:
Thanks, makes sense!

